PR #19, opened 2016-03-12
implement fmt::Display for Matrix

zackmdavis:
Users of a linear algebra library might occasionally want to pretty-print a matrix!

``` rust
use rusty_machine::linalg::matrix::Matrix;

fn main() {
    let m = Matrix::new(3, 3, (1..10).collect::<Vec<_>>());
    println!("{}", m);
}

// outputs:
//
// ⎡1 2 3⎤
// ⎢4 5 6⎥
// ⎣7 8 9⎦
```

### DIFF
--- a/rusty-machine/src/linalg/matrix/mod.rs
+++ b/rusty-machine/src/linalg/matrix/mod.rs
@@ -3,6 +3,7 @@
 //! Currently contains all code
 //! relating to the matrix linear algebra struct.
 
+use std::fmt;
 use std::ops::{Mul, Add, Div, Sub, Index, Neg};
 use libnum::{One, Zero, Float, FromPrimitive};
 use std::cmp::{PartialEq, min};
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Another super useful contribution, thank you!

Before I merge this one I'd like to get your opinion. You added the test to an internal `tests` module and I've seen this a lot through other repos. Do you think it is worth migrating the tests I currently have in the `tests` folder to internal modules as well? If so I'll make an issue for that.

My current opinion is that it probably is useful - it's hard maintaining those separate tests right now (and they need updating!).

AtheMathmo:
I looked a little into display accuracy. We can control precision based on the print statement, e.g.

```
use rusty_machine::linalg::matrix::Matrix;

fn main() {
    let m = Matrix::new(2, 2, [1.234, 2.3456, 1.1, 2.567]);
    println!("{:.2}", m);
}

// should output:
//
// ⎡1.23 2.34⎤
// ⎢1.10 2.56]
```

To achieve this we use the following in the fmt function:

```
if let Some(prec) = f.precision() {
    try!(write!(f, "{:>2$.*}", prec, datum, width));
}
```

This will require computing the width differently (if a precision is present we will have to compute it using this precision.

It's your call as to whether this is included in this PR - I'm happy to write up a separate ticket.

zackmdavis:
> You added the test to an internal tests module and I've seen this a lot through other repos. Do you think it is worth migrating the tests I currently have in the tests folder to internal modules as well?

The [relevant sections of The Book](https://doc.rust-lang.org/book/testing.html#the-tests-module) seem to recommend same-file submodules for unit tests and a tests directory for larger-scale integration tests (exercising multiple features of the library the way a client would).

> Do you think it is possible to set a display accuracy [...] This will require computing the width differently

Good point; what do you think of 865bc6dc1? (fa221ad0 also adds a test for the one-row special case.)

AtheMathmo:
The new changes are great; thanks for your work! I'll release a new version with these improvements soon.

PR #21, opened 2016-03-13
Derive the World

zackmdavis:
Herein:

* remaining `derive(Debug)` attributes, a sequel to #18  
* activating the missing-`Debug` lint at `warn` level
* `derive(Copy, Clone)` attributes where applicable ("[Generally speaking,](https://doc.rust-lang.org/std/marker/trait.Copy.html#when-should-my-type-be-copy) if your type can implement Copy, it should.")


AtheMathmo:
Thank you again, these changes are really helpful. I hope you can find more improvements like these!

Issue #26, opened 2016-03-19
Reallocating memory when consuming matrix in operations


AtheMathmo:
### The current situation

When we consume matrices during operations we are allocating new memory to contain the result.

e.g.

```
let a = Matrix::new(3,3, vec![2.0; 9]);
let b = Matrix::new(3,3, vec![2.0; 9]);

// We deallocate the memory for a and b, but allocate new memory for c.
let c = a + b;
```

### Suggested solution

We can reuse the allocated memory for the consumed matrix. This means mutating the matrix to be consumed.

This will mean that we do not need to allocate new memory.

### Affected code

We can make this change for all of the standard operations except for multiplication. In the case where we are multiplying two square matrices we can also make the same change.

PR #27, opened 2016-03-20
Memory alloc

AtheMathmo:
This solves (largely) issue #26 . It also includes some improvements in terms of vectorization and code re-usability.

The PR includes changes to the underlying binary data operations (sum, subtract, element-wise multiplication, etc.). It also reuses memory when one of the incoming arguments is to be consumed (and has the appropriate size).

One aspect ignored in this PR is the reassignment of matrix data during square matrix multiplication. I figured this is a rare case in the problem domain but should be introduced later.

#### Benches

test linalg::matrix::empty                                 ... bench:               0 ns/iter (+/- 0)
test linalg::matrix::mat_create_100_100         ... bench:        2,734 ns/iter (+/- 254)
test linalg::matrix::mat_create_add_100_100 ... bench:        8,113 ns/iter (+/- 90)
test linalg::matrix::mat_mul_10_10                 ... bench:           283 ns/iter (+/- 5)
test linalg::matrix::mat_old_add_100_100      ... bench:      23,205 ns/iter (+/- 2,158)
test linalg::matrix::mat_ref_add_100_100       ... bench:        6,725 ns/iter (+/- 105)

We can see huge improvements from just the vectorization alone. The new matrix add takes on average 6725 ns/iter compared to the 23,000 ns/iter before. 

We also have a small gain from reusing memory. If we subtract the creation time it costs us an average of 5400 ns/iter to add matrices when we reuse the memory. This gain will be more significant with larger sized matrices (which is common in our domain).

### DIFF
--- a/rusty-machine/src/linalg/matrix/mod.rs
+++ b/rusty-machine/src/linalg/matrix/mod.rs
@@ -363,14 +363,12 @@ impl<T: Copy> Matrix<T> {
     ///
     /// assert_eq!(*b.data(), vec![2.0; 4]);
     /// ```
-    pub fn apply(self, f: &Fn(T) -> T) -> Matrix<T> {
-        let new_data = self.data.into_iter().map(f).collect();
-
-        Matrix {
-            rows: self.rows,
-            cols: self.cols,
-            data: new_data,
+    pub fn apply(mut self, f: &Fn(T) -> T) -> Matrix<T> {
+        for val in self.data.iter_mut() {
+            *val = f(*val);
         }
+
+        self
     }
 }
 
@@ -1251,8 +1249,8 @@ impl<'a, 'b, T: Copy + One + Zero + Sub<T, Output = T>> Sub<&'b T> for &'a Matri
 impl<T: Copy + One + Zero + Sub<T, Output = T>> Sub<Matrix<T>> for Matrix<T> {
     type Output = Matrix<T>;
 
-    fn sub(self, f: Matrix<T>) -> Matrix<T> {
-        self - &f
+    fn sub(self, m: Matrix<T>) -> Matrix<T> {
+        self - &m
     }
 }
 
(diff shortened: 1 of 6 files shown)


bluss:
What did the benches say? That's always the best part! I like using https://github.com/BurntSushi/cargo-benchcmp to compare two benchmark result logs.

rusty-machine/src/linalg/utils.rs
```diff
@@ -111,6 +112,64 @@ pub fn unrolled_sum<T>(mut xs: &[T]) -> T
...
+pub fn in_place_vec_bin_op<F, T: Copy>(mut u: &mut [T], v: &[T], mut f: F)
+    where F: FnMut(&mut T, &T) {
+        debug_assert_eq!(u.len(), v.len());
+        let len = cmp::min(u.len(), v.len());
+        
+        let ys = &v[..len];
+        
+        let mut new_slice = mem::replace(&mut u, &mut []);
+        let (l,_) = new_slice.split_at_mut(len);
+        u = l;
```
bluss: This can be simplified to just `let xs = &mut u[..len]` (and adjust the use of `u` down the code).

rusty-machine/tests/linalg/mat.rs
```diff
@@ -138,6 +170,37 @@ fn matrix_add() {
     let a = Matrix::new(3, 2, vec![1., 2., 3., 4., 5., 6.]);
     let b = Matrix::new(3, 2, vec![2., 3., 4., 5., 6., 7.]);
 
+    // Reusing memory
+    let c = &a + &b;
```
krdln: This should be "allocating", right?

rusty-machine/src/linalg/matrix/mod.rs
```diff
@@ -1128,8 +1130,12 @@ impl<'a, T: Copy + One + Zero + Add<T, Output = T>> Add<T> for &'a Matrix<T> {
...
     type Output = Matrix<T>;
 
-    fn add(self, f: &T) -> Matrix<T> {
-        (&self) + f
+    fn add(mut self, f: &T) -> Matrix<T> {
+        for val in self.data.iter_mut() {
+            *val = *val + *f;
+        }
+
+        self
```
krdln: You could use your `apply` function here (and in other `Matrix` op scalar functions)

rusty-machine/src/linalg/utils.rs
AtheMathmo: Thanks! I guess this was a remnant of when this code looked quite different.

rusty-machine/tests/linalg/mat.rs
AtheMathmo: Should indeed, thank you!

rusty-machine/src/linalg/matrix/mod.rs
AtheMathmo: You're right, I'll make that change too - I don't think it makes the code any harder to follow.

AtheMathmo:
Added some benchmarks to the overview above.

bluss:
nice

Issue #29, opened 2016-03-25
Cleanup the operation overloading


AtheMathmo:
#### What is the problem?

Right now there are ~500 lines of code in `matrix/mod.rs` dedicated to operation overloading. The new `matrix/slice.rs` module will add another ~500 (or more). This consists largely of repeated code for different cases.

#### What should we do about it?

This issue should track the investigation and implementation of how we can clean up the operation overloading. There are two solutions clear to me:

1. Use macros to cut down on repetitive code.
2. Move the operation overloading into new modules.

It is likely that a combination of the above would be best. The `matrix/mod.rs` file is 1500 lines long so should really be cut down.

Any other suggestions would be welcomed.

AtheMathmo:
This post will provide a summary of the operators that we need to overload. It will hopefully make it a little easier to plan out which macros we should implement.

### Summary of operators:

**Matrix binary op Scalar**

Two variants.

- If `Matrix` is consumed, we iterate:

```rust
fn mul(mut self, f: &T) -> Matrix<T> {
   for val in self.data.iter_mut() {
       *val = *val * *f;
    }

    self
}
```

- If `Matrix` is not consumed we iterate and map:

```rust
fn mul(self, f: &T) -> Matrix<T> {
    let new_data: Vec<T> = self.data.iter().map(|v| (*v) * (*f)).collect();

    Matrix {
        cols: self.cols,
        rows: self.rows,
        data: new_data,
    }
}
````

**Matrix binary op Matrix**

We have two variations again - and sub-variations where our operation is non-commutative.

- If one of the `Matrix` structs is consumed:

```rust
fn add(mut self, f: &Matrix<T>) -> Matrix<T> {
    utils::in_place_vec_bin_op(&mut self.data, &f.data, |x,&y| {*x = *x + y});
    self
}
```

- If neither `Matrix` is consumed:

```rust
fn add(self, m: &Matrix<T>) -> Matrix<T> {
    // We could also use `vec_bin_op` which is itself called within `vec_sum`.
    let new_data = utils::vec_sum(&self.data, &m.data);

    Matrix {
        cols: self.cols,
        rows: self.rows,
        data: new_data,
    }
```

- If we are subtracting matrices we must use different operations when matrices are consumed. This should be solvable with a single matrix - we swap the order by default which will not affect commutative operations.

**Matrix \* Matrix**

Currently we always allocate new memory (the only case we can reuse memory is with square matrices. This boils down to a loop like this:

```rust
[...]

Issue #22, opened 2016-03-16
Implementing Matrix Views


AtheMathmo:
### Preface

It is common in numerical linear algebra to do operations on chunks of a matrix. This is common within rusty-machine. We often need to select the top `n` rows of a matrix, and similar other operations.

### Use Case

The following use case is one that is not currently in the library - however will be soon. We want to implement Parallel Divide and Conquer matrix multiplication. In order to do this at each step we must split our matrix into chunks and multiply these smaller pieces together.

### Current Situation

Currently to do this we are utilizing the `select_rows`, `select_cols` and `select` matrix functions. These functions take a reference to a `Matrix` and copy the rows/columns specified into a new `Matrix`. In some situations this could mean large amounts of data being copied into new `Matrix` structs.

### Proposal

We should implement a `MatrixView` struct (name pending). This struct will contain:

- A pointer to the underlying matrix.
- An array `[usize; 2]`, specifying the top left point of the view.
- A row size, specifying the width of the view.
- A column size, specifying the height of the view.

This will allow us take a chunk of the matrix without copying large amounts of data. We can implement the various arithmetic and indexing traits on this `MatrixView` and will be able to use it in place of `Matrix` in some situations.

### Notes

- We will still require the various `select` methods in many places for `Matrix` directly.
- We should not be able to directly instantiate a `MatrixView`. We should be able to create this from a Matrix, e.g.

```
pub fn from_matrix(mat: &Matrix<T>, start: &[usize;2], rows: usize, cols: usize) -> MatrixView<T>
```

- We should provide a similar `select` method for the `Matrix` struct which produces the view.
- The view should also be able to be `select`ed from (relative to the view). This would mean producing a new `MatrixView` with the same underlying matrix.
- The view should be contained within it's own module. The `mod.rs` file for `matrix` is already pretty bloated...
- The various arithmetic operations will produce `Matrix` structs.

### Open Questions

- Can we provide efficient iterators using the view?
- Please comment below with any questions on details or suggestions for improvement.

AtheMathmo:
The implementation I've detailed above has a lot of code repetition with the arithmetic. We can work around this with a common matrix trait - but I'm not sure if this is worth the extra complexity.

For now I'll continue implementing the arithmetic as is.

PR #30, opened 2016-03-27
Add setters and a specified constructor: kmeans

danlrobertson:
Add a constructor with specified values for all options for the KMeansClassifier. In addition, add setters for each field. Not super attached to any of the components of the patch besides `KMeansClassifier::new_specified`. I also added some rules to `.gitignore` for the sanity of vim users :smile:

---
This change is [Reviewable](https://reviewable.io/reviews/athemathmo/rusty-machine/30)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 # compiled code
 rusty-machine/target/
 rusty-machine/Cargo.lock
+*~
+*.swp
+*.swo
--- a/rusty-machine/src/learning/k_means.rs
+++ b/rusty-machine/src/learning/k_means.rs
@@ -60,6 +60,93 @@ pub enum InitAlgorithm {
     KPlusPlus,
 }
 
+/// K-Means Classification Builder.
+#[derive(Debug)]
+pub struct KMeansClassifierBuilder {
+    k: usize,
+    iters: usize,
+    init_algorithm: InitAlgorithm
+}
+
+impl KMeansClassifierBuilder {
+    /// Create a K-Means Classification Builder.
+    ///
+    /// Requires number of classes to be specified.
+    /// Defaults to 100 iterations and kmeans++ initialization.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rusty_machine::learning::k_means::KMeansClassifierBuilder;
+    ///
+    /// // Creates a new KMeansClassifier with 5 clusters
+    /// let model = KMeansClassifierBuilder::new(5).finalize();
+    /// ```
+    pub fn new(k: usize) -> Self {
+        KMeansClassifierBuilder {
+            k: k,
+            iters: 100,
+            init_algorithm: InitAlgorithm::KPlusPlus
+        }
+    }
+
+    /// Changes the number of iterations.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rusty_machine::learning::k_means::KMeansClassifierBuilder;
+    ///
+    /// // Creates a new KMeansClassifier with 5 clusters and 42 iterations
+    /// let model = KMeansClassifierBuilder::new(5).iters(42).finalize();
+    /// ```
+    pub fn iters(&mut self, iters: usize) -> &mut KMeansClassifierBuilder {
+        self.iters = iters;
+        self
+    }
+
+    /// Changes the initialization algorithm.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rusty_machine::learning::k_means::KMeansClassifierBuilder;
+    /// use rusty_machine::learning::k_means::InitAlgorithm;
+    ///
+    /// // Creates a new KMeansClassifier with 5 clusters and uses
+    /// // the Forgy initialization algorithm
+    /// let model = KMeansClassifierBuilder::new(5).init_algorithm(InitAlgorithm::Forgy)
+    ///                                            .finalize();
+    /// ```
+    pub fn init_algorithm(&mut self, init_algorithm: InitAlgorithm) -> &mut KMeansClassifierBuilder {
+        self.init_algorithm = init_algorithm;
+        self
+    }
+
+    /// Returns the KMeansClassifier with the options previously specified.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rusty_machine::learning::k_means::KMeansClassifierBuilder;
+    /// use rusty_machine::learning::k_means::InitAlgorithm;
+    ///
+    /// // Creates a new KMeansClassifier with 5 clusters, 42 iterations
+    /// // using the Forgy initialization algorithm
+    /// let model = KMeansClassifierBuilder::new(5).iters(42)
+    ///                                            .init_algorithm(InitAlgorithm::Forgy)
+    ///                                            .finalize();
+    /// ```
+    pub fn finalize(&self) -> KMeansClassifier {
+        KMeansClassifier {
+            k: self.k,
+            iters: self.iters,
+            centroids: None,
+            init_algorithm: self.init_algorithm
+        }
+    }
+}
+
 /// K-Means Classification model.
 ///
 /// Contains option for centroids.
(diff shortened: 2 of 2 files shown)


AtheMathmo:
Thanks for helping out! K-Means was one of the first algorithms I implemented and so was some of the first rust code I wrote - it definitely needs some work.

As of right now all of the struct fields are public and so setter's aren't _needed_. However, I agree with your logic that we probably shouldn't allow fields to be set on models which have been trained. I think we should make the struct fields private and introduce these setters. This is a breaking change though and so I may hold off on merging it for a little while.

Also I totally agree there should be a constructor that doesn't just use the default initialization algorithm. I'm not a huge fan of the name you picked but I can't think of anything better :D.

_Irrelevant to PR: Would be great to see some of this macro magic on issue #29 if yo're up for it? It's a much bigger task though and also should probably be tackled after #22. No pressure!_

danlrobertson:
> I think we should make the struct fields private and introduce these setters. This is a breaking change though and so I may hold off on merging it for a little while.
>
> Also I totally agree there should be a constructor that doesn't just use the default initialization algorithm. 

I agree. My current thoughts... So perhaps this should be broken up into 2 PRs. One to add another constructor (sorry for the C++isms :smile:) and one large PR probably maintained on a separate branch to be merged at some later date with a major or minor version bump (depending on how you're going to handle releases and versioning) that uses getters and setters for algorithm fields (ensuring a common interface across all algorithm structures).

> I'm not a huge fan of the name you picked but I can't think of anything better :D.

I am absolutely horrendous at naming things :smile:

> Irrelevant to PR: Would be great to see some of this macro magic on issue #29 if yo're up for it? It's a much bigger task though and also should probably be tackled after #22. No pressure!

I'd be happy to take a look. I've got a few other PRs && isssues that I've got to do some work on, but I'll take a look and post my findings/thoughts on the issue sometime this week

AtheMathmo:
I think you're suggestion is a good one. I'll merge the changes you propose in this PR and include them as part of the next patch release. We'll include the larger changes in a separate PR.

Currently my plan is to allow (minor) breaking changes in minor (0.X.0) releases. For example making the k-means fields private but providing setters. Patches (0.0.X) releases should be completely back compatible (though I have been adding functionality in these versions too).

I'll spend a couple days mulling over the name (I'm also terrible at naming things, as you can probably tell). It will likely remain as you suggested though.

Thanks again!

zackmdavis:
On the question of argument signature for initializers where some parameters should have a default, the [builder pattern](https://doc.rust-lang.org/book/method-syntax.html#builder-pattern) might be a good fit?

``` rust
struct KMeansClassifierBuilder {
    k: usize,
    iters: usize,
    init_algorithm: InitAlgorithm
}

impl KMeansClassifierBuilder {
    fn new(k: usize) -> Self {
        KMeansClassifierBuilder {
            k: k,
            iters: 100,
            init_algorithm: InitAlgorithm::KPlusPlus
        }
    }

    fn iters(&mut self, iters: iters) -> &mut KMeansClassifierBuilder {
        self.iters = iters;
        self
    }

    // ... &c.

    fn finalize(&self) -> KMeansClassifier {
        KMeansClassifier {
            k: k,
            iters: 100,
            centroids: None,
            init_algorithm: InitAlgorithm::KPlusPlus
        }
    }
}
```

AtheMathmo:
Thanks for checking out the PR Zack!

I'm at a bit of a crossroads with this one. I really like the builder syntax in rust but there are a couple issues with implementing it here (I think):

- We don't use it anywhere else in the library. I'm trying to focus on homogeneity within the library and this goes against that.
- This is a very **rust**y thing to do. I _think_ most of the users of this library will be either new to Rust or new to machine learning and I don't want to isolate one group or the other.

The first issue could be solved by using this syntax throughout the library. The second I'm not sure can be solved?

These are just my thoughts though. I'm happy to be overruled by popular opinion and make this change in a future larger update.

AtheMathmo:
@danlrobertson @zackmdavis - Do either of you have any more thoughts on this PR? 

I'm still pretty conflicted but I think if there are no objections I'll merge the PR as is - it adds some needed functionality without breaking anything. We'll definitely need to address the issues discussed initially but can create a new issue for that.

I do like the builder pattern - especially in this case. But I'm not sure how it affects the balance I discussed in the previous comment.

danlrobertson:
I definitely think the Builder pattern is the way to go, but I agree that it may cause less experienced users some hardships. My thoughts...

 - Add a constructor that allows setting everything (mostly done)
 - Make the current members private
 - Add `setters` and `getters` (mostly done)
 - @zackmdavis submits a PR with his example code using the builder pattern and highly publicize this as the way to go in any documentation and tutorials, but hang on to the current API to allow for easy use by newer users.

Just my thoughts... I'm happy to do some work on this PR if there is consensus on the above or other items.

AtheMathmo:
Ok, I think we should go with the builder syntax. However if we do this I think to keep a clean API we should not have setters on the KMeansClassifier struct (this is just my thinking, if this goes against convention let me know!).

I would modify your suggestion to the following:

- Keep the current _default_ constructor which lets us choose only the number of clusters. I think this should be available to keep things easy as you say - and additionally the settings aren't critically important for K-Means.
- Make the current members private. (Perhaps keep them public for this initial PR, and have a new PR to make them private as a breaking change).
- Add `getters`.
- Add the builder to this PR. Include example usage in module documentation, and for the struct documentation.

This to me seems like a good compromise - we don't break things (yet). We provide the right way to build a k-means classifier. We keep things rust-like without adding a necessary learning curve.

How does this sound to each of you? I'm happy again to be overruled.

zackmdavis:
_(shrugging)_ My intuition kind of expects `::new` to take the complete complement of possible arguments, but I understand that that would have the non-ergonomic consequence of the I-don't-care-just-give-me-the-defaults initializer call being `KMeansClassifierBuilder::new(k).finalize()`, which we likely want to avoid ...

AtheMathmo:
You make a good point. It seems that due to my current shoddy design that whatever has some negative consequence.

I think we should still have a simple default constructor for the `KMeansClassifier` but I'm not too concerned about how this looks exactly. In the world of idiomatic Rust would you expect to have a fully specified `new` function if there is a builder present as well? We could always change the `new` function to `from_cluster_count` or something less verbose?

danlrobertson:
@zackmdavis: I made a commit using your code and committed using you as the author. Let me know if you have any issues with that. If you would rather submit a separate PR or a PR against my branch. I'm totally okay with that. I just thought this was probably the easiest option.

Still not perfect, but I think we're getting there. After writing out the examples I'm really starting to see the benefit of going the route of using a builder pattern.

zackmdavis:
> using you as the author. Let me know if you have any issues with that.

I appreciate the gesture, but I think the commit authorship should be in your name; my comment with the trivial code snippet was just a comment (which might or might not be worth citing in the commit message, but IMO isn't worth forging the author field for).

AtheMathmo:
I agree that this is looking a lot better - the builder pattern does indeed seem like the way to go. I'm still unsure about including the `new_specified` method in addition to the builder - it is redundant and I feel that we will want to remove it in the future. That said - it isn't bad to include it at this point.

Another minor point - my understanding is that in Rust it is standard to exclude the `get_` part on your get methods, so the methods in this PR would become:

```rust
/// Get the number of classes
pub fn k(&self) -> usize {
    self.k
}
 
/// Get the number of iterations
 pub fn iters(&self) -> usize {
    self.iters
}
 
/// Get the initialization algorithm
pub fn init_algorithm(&self) -> InitAlgorithm {
    self.init_algorithm
}
```

danlrobertson:
Awesome! I'm also conflicted about the name, but I can't think of anything better...

I squashed the two commits and removed the `get_` from the getters.

AtheMathmo:
My argument for `new_specified` was more that we can now achieve this functionality using the builder and so no longer need this function at all. However it may keep things easy if we include it for now :).

I just checked out the branch and there a couple provlems. It looks like there is a typo, on line 100:

```rust
fn iters(&mut self, iters: iters) ...
```

has `iters` as the type. It should read:

```rust
fn iters(&mut self, iters: usize) ...
```

Additionally the builder methods need to be public or they cannot be accessed (the current doc tests fail). We also do not need to return the `&mut` reference (although the builder example @zackmdavis gave does). I'd suggest the function should be, e.g:

```rust
pub fn iters(&mut self, iters: usize) {
    self.iters = iters;
}
```

Are you happy making these changes?

AtheMathmo:
The functionality here is looking great now - thanks for your hard work!

There is only one more thing I have just thought of - which I am happy to exclude (you've already put up with a lot of my comments!). I think we should add some documentation explaining what the purpose of the builder is. I'm thinking that the `KMeansClassifierBuilder` struct documentation should explain what it does? I think the usage is described well enough in the examples (though it may be worth adding some inline comments to make it clear that we get a `KMeansClassifier` out. e.g.

/// ```
/// use rusty_machine::learning::k_means::KMeansClassifierBuilder;
///
/// // Create a new classifier with 5 clusters.
/// let model : KMeansClassifier = KMeansClassifierBuilder::new(5).finalize();
/// ```

As I say though - not necessary. Give me the go ahead and I'll merge. Thanks again!

danlrobertson:
Thanks for the comments. They were super helpful! Sorry it took so many revisions :smile:

AtheMathmo:
Glad it was helpful :). And I think it took so many revisions as we were figuring what we needed as we went along - thanks for your help delivering a great addition!

PR #41, opened 2016-04-15
Cleaning up operator overloading

AtheMathmo:
This resolves issue #29 . We will likely need more work in the future but this is a big improvement.

I have moved all of the operator overloading for `Matrix`, `MatrixSlice` and `MatrixSliceMut` into a new module, `impl_ops`. This makes the functionality for the struct modules much easier to understand (I think).

I've swapped out the current matrix implementations for macros - which should make things a lot easier to deal with.


AtheMathmo:
@danlrobertson @zackmdavis - I want to merge this one pretty fast so I can move on to things that depend on it. If either of you have time to take a look I'd appreciate it!

It's a pretty big PR so feel free to ignore :)

rusty-machine/src/linalg/matrix/impl_ops.rs
```diff
@@ -0,0 +1,648 @@
...
+    fn neg(self) -> Matrix<T> {
+        let new_data = self.data.iter().map(|v| -*v).collect();
+
+        Matrix {
+            cols: self.cols,
+            rows: self.rows,
+            data: new_data,
+        }
+    }
+}
```
zackmdavis: You can probably configure your error to automatically put a newline at end of files so that diff tools won't complain like this.

zackmdavis: You can probably configure your error to automatically put a newline at end of files so that diff tools won't complain like this.

PR #43, opened 2016-04-16
Matrixmultiply

### LINKED ISSUE #38: matrixmultiply
[matrixmultiply](https://github.com/bluss/matrixmultiply) does f32 and f64 matrix multiplication quite competently for larger matrices. It's a simpler dependency than blas, so it might be nice to just use. It does need type specific dispatch (unstable stabilization, TypeId, or specific trait) at this point though.

What do you think, could I help you integrate with it, or are you developing something equivalent?

AtheMathmo:
This gives us huge speedups for matrix multiplication:

> Before:
> 
> test linalg::matrix::mat_mul_128_1000 ... bench: 14,231,183 ns/iter (+/- 833,129)
> 
> After:
> 
> test linalg::matrix::mat_mul_128_1000 ... bench: 3,352,754 ns/iter (+/- 412,408)

This change only applies for f32 and f64 multiplication. The PR adds some more complex trait bounds (requiring `Any` on many implementation blocks). Once specialization lands this can be removed.

This resolve #38 .

### DIFF
--- a/rusty-machine/src/linalg/matrix/decomposition.rs
+++ b/rusty-machine/src/linalg/matrix/decomposition.rs
@@ -5,6 +5,7 @@
 
 use std::ops::{Mul, Add, Div, Sub, Neg};
 use std::cmp;
+use std::any::Any;
 
 use linalg::matrix::Matrix;
 use linalg::vector::Vector;
@@ -14,7 +15,7 @@ use linalg::utils;
 use libnum::{One, Zero, Float, NumCast, Signed};
 use libnum::{cast, abs};
 
-impl<T: Copy + Zero + Float> Matrix<T> {
+impl<T: Any  + Float> Matrix<T> {
     /// Cholesky decomposition
     ///
     /// Returns the cholesky decomposition of a positive definite matrix.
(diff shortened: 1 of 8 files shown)


rusty-machine/src/linalg/matrix/decomposition.rs
```diff
@@ -159,7 +160,7 @@ impl<T: Copy + Zero + Float> Matrix<T> {
     }
 }
 
-impl<T: Copy + Zero + One + Float + NumCast + Signed> Matrix<T> {
+impl<T: Any + Copy + Zero + One + Float + NumCast + Signed> Matrix<T> {
```
bluss: I Copy, Zero, One, NumCast are all redundant here (included in Float, if that's num::Float).

rusty-machine/src/linalg/matrix/mat_mul.rs
```diff
@@ -0,0 +1,243 @@
...
+            let mut new_data = Vec::with_capacity(p * r);
+            
+            unsafe {
+                new_data.set_len(p * r);
+
+                matrixmultiply::sgemm(
+                    p, q, r,
+                    1f32,
+                    self.as_ptr() as *const _,
+                    q as isize, 1,
```
bluss: I forgot -- you have MatrixSlices here too. Then you must use their row stride.

AtheMathmo: You're totally right! I'll have to figure out why my tests didn't catch this...

PR #45, opened 2016-04-17
Using row-stride with matrixmultiply

AtheMathmo:
Fixing bug with MatrixSlice multiplication. Adding more test coverage.


bluss:
:+1: :1234:  :runner:

Issue #48, opened 2016-04-20
Adding row iterators


AtheMathmo:
It would be nice to be able to iterate over the rows in our matrices.

This will also allow us to make some optimizations for binary operation assignment on `MatrixSliceMut`. We cannot (easily) vectorize the whole operation as the data is not contiguous - but the rows are. If we can iterate over the rows then we can vectorize the operation on each row.

zackmdavis:
Can we get column iterators, too?

(Last night, when I was starting to write the eigenvectors test for #47 before I ran out of time, I found myself wanting to iterate over the columns of the matrix that we return from `eigendecomp` so that I could assert that for each eigenvalue λ_i and corresponding eigenvector v_i, A*v_i = λ_i * v_i, which seemed more meaningful than asserting that the eigenvector is componentwise identical to some expected value, but I didn't see a column iterator anywhere.)

AtheMathmo:
Column iterators are less easy (at least to do efficiently). Rows are ok because we store the matrix in row-major order - we don't need to copy any of the data. To do a column iterator we need to be able to copy the data and place it into contiguous slices.

I do agree though that it would probably be good to have despite this. I'll see if I can figure out a nice way to do it :). We could even be hacky and do an in-place transpose (which I still need to write) and then return a Row iterator...

PR #49, opened 2016-04-20
Adding row iterators for each matrix struct.

AtheMathmo:
This is to resolve issue #48 .

### DIFF
--- a/rusty-machine/src/linalg/matrix/iter.rs
+++ b/rusty-machine/src/linalg/matrix/iter.rs
@@ -1,0 +1,456 @@
+use std::iter::{ExactSizeIterator, FromIterator};
+use std::marker::PhantomData;
+use std::slice;
+
+use super::{Matrix, MatrixSlice, MatrixSliceMut};
+
+/// Row iterator.
+#[derive(Debug)]
+pub struct Rows<'a, T: 'a> {
+    slice_start: *const T,
+    row_pos: usize,
+    slice_rows: usize,
+    slice_cols: usize,
+    row_stride: isize,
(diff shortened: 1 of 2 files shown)


rusty-machine/src/linalg/matrix/iter.rs
```diff
@@ -0,0 +1,194 @@
...
+    row_pos: usize,
+    slice_rows: usize,
+    slice_cols: usize,
+    row_stride: isize,
+    _marker: PhantomData<&'a T>,
+}
+
+/// Iterates over the matrix data by slice
+impl<'a, T> Iterator for Rows<'a, T> {
+    type Item = &'a [T];
```
zackmdavis: I think I could imagine wanting the iterator elements to each be a `MatrixSlice<T>` or a `Vector<T>` rather than a `&[T]`? I'm not sure.

AtheMathmo: `MatrixSlice<T>` may be a good idea actually - I hadn't really considered it before. I wouldn't use a `Vector<T>` as this needs to own it's data (it uses a `Vec`).

The reason I wrote it right now to return slices is that I want to use this iterator to provide more efficient OpAssign methods for `MatrixSliceMut`. This will be helpful/needed for the multithreaded matrix multiplication work.

AtheMathmo:
I realised we can also add a FromIterator trait which would work nicely with this approach. For example see this [playground code](https://play.rust-lang.org/?gist=4bf1c947206f5215c84876a04c084be1&version=stable&backtrace=0).

PR #50, opened 2016-04-21
derive PartialEq for Matrix and Vector

zackmdavis:
Any reason not to do this?

(Previously—

```
<std macros>:5:8: 5:18 error: binary operation `==` cannot be applied to type `linalg::matrix::Matrix<_>` [E0369]
<std macros>:5 if ! ( * left_val == * right_val ) {
                      ^~~~~~~~~~
src/linalg/matrix/mod.rs:1190:9: 1190:32 note: in this expansion of assert_eq! (defined in <std macros>)
<std macros>:5:8: 5:18 help: run `rustc --explain E0369` to see a detailed explanation
<std macros>:5:8: 5:18 note: an implementation of `std::cmp::PartialEq` might be missing for `linalg::matrix::Matrix<_>`
<std macros>:5 if ! ( * left_val == * right_val ) {
                      ^~~~~~~~~~
src/linalg/matrix/mod.rs:1190:9: 1190:32 note: in this expansion of assert_eq! (defined in <std macros>)
```
)

### DIFF
--- a/rusty-machine/src/linalg/matrix/mod.rs
+++ b/rusty-machine/src/linalg/matrix/mod.rs
@@ -29,7 +29,7 @@ pub enum Axes {
 /// The Matrix struct.
 ///
 /// Can be instantiated with any type.
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub struct Matrix<T> {
     rows: usize,
     cols: usize,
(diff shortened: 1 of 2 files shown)


AtheMathmo:
I see no reason not to do this! I think in practice we may want some helper `approx_eq` function for comparing `Matrix<T : Float>`. That's just a thought though - I'll merge this as is.

Thanks!

PR #53, opened 2016-04-22
Matrix and Vector constructors accept data as Into<Vec<T>>?

zackmdavis:
Arguments in favor:
  * Flexibility! You have the option of passing a slice (or anything else that implements `Into<Vec<T>>`) and letting `new` allocate its own vector.

Arguments against:
  * type inference doesn't work as well: notice how this commit had to add several turbofish operators to tests.
    * rebuttal: actual users mostly aren't going to be getting their data from `.collect()`, right?

### DIFF
--- a/rusty-machine/src/linalg/matrix/impl_ops.rs
+++ b/rusty-machine/src/linalg/matrix/impl_ops.rs
@@ -1115,7 +1115,7 @@ mod tests {
 
     #[test]
     fn index_slice() {
-        let mut b = Matrix::new(3, 3, (0..9).collect());
+        let mut b = Matrix::new(3, 3, (0..9).collect::<Vec<_>>());
 
         let c = MatrixSlice::from_matrix(&b, [1, 1], 2, 2);
 
@@ -1187,7 +1187,7 @@ mod tests {
     #[test]
     fn matrix_sub_assign() {
         let mut a = Matrix::new(3, 3, (0..9).collect::<Vec<i32>>());
-        
+
         a -= &2;
         assert_eq!(a.into_vec(), (-2..7).collect::<Vec<_>>());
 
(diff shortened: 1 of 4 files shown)


AtheMathmo:
I'm a little conflicted about naming here but I love the functionality. I do agree with your rebuttal - I only ever do this to generate some non-constant data easily in the tests.

I _think_ (see note after) I will hold off on merging this until a few more breaking changes come in for 0.3.0. I am planning on writing up an issue to discuss which breaking changes we'd like to make in a new minor version bump.

_Note: this is a pretty minor breaking change so I may just sneak it in as 0.2.X ... What are your thoughts?_

zackmdavis:
Sure, waiting for 0.3 is good semantic-versioning citizenship.

(Although in the absence of actual users clamoring for features and fixes, the _pragmatic_ reasons for carefully considering what goes in 0.(X+1).0 vs. 0.X.Y are greatly attenuated; in the wild west greenfield days at the beginning of the world, many projects can get away with letting everything hit master as soon as it seems to constitute an improvement, occasionally tagging 0.N.0 more for the sake of advertising to hypothetical users everything that happened since the last tag, than for the sake of protecting actual users from breaking changes.)

AtheMathmo:
@zackmdavis - sorry I threw some other changes in and produced a conflict. Totally my fault...

Would you mind fixing the conflict? If you're busy I can do it. Sorry again!

zackmdavis:
(rebased)

> a little conflicted about naming here

I believe that as project creator and BDFL, you have a moral obligation to ruthlessly enforce whatever naming conventions you prefer (which mere contributors may of course _question_, but with the understanding that after everything has been said, there can be no higher recourse than the freedom to fork).

AtheMathmo:
Ha! I will certainly do my best to ensure ruthless conventions in future. In this situation I decided your name was probably the most idiomatic way to handle the change, `new` is about as appropriate as anything else.

Thank you for your continued work - it's made a huge difference already!

PR #57, opened 2016-04-28
clean up compiler warnings

zackmdavis:
The augmented-assignment tests in particular generated a lot of
unused-variable and unused-assignment warnings, which are undesirable
not only on æsthetic developer-experience grounds but, more importantly,
because genuinely useful warnings inadvertently set off by future
development efforts could get lost in the noise. Typically one would
indicate an intentionally-unused variable by choosing an identifier
starting with an underscore, but in the case of the augmented-assignment
tests, the `allow`-directive attributes seem more appropriate: the
`a_slice` variables weren't _really_ unused; it's just that `rustc`
isn't smart enough to notice that they're views into an underlying
matrix which is being used.

### DIFF
--- a/rusty-machine/src/linalg/matrix/impl_ops.rs
+++ b/rusty-machine/src/linalg/matrix/impl_ops.rs
@@ -1187,7 +1187,7 @@ mod tests {
     #[test]
     fn matrix_sub_assign() {
         let mut a = Matrix::new(3, 3, (0..9).collect::<Vec<i32>>());
-        
+
         a -= &2;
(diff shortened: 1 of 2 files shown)


AtheMathmo:
I haven't been getting these compiler warnings that you're seeing on the op assignment tests. I do however see the `output` warning which I have been meaning to fix...

I'm happy to accept this fix but want to try and figure out the reason why it is occurring first!

Could you confirm which version of Rust you're using? I'm on 1.8.0.

PR #72, opened 2016-05-15
Into<Matrix> and From<>-for-Matrix, for Vector and matrix slices

zackmdavis:
`std::convert` provides slick `From` and `Into` traits for converting between types (and implementing `From` automatically gives you `Into`); we could be making more use of them (in addition to accepting U: Into<Vec<T>> as initialization data, as seen in #53) in the following two ways—

* By implementing `From<Vector<T>>` for `Matrix<T>`, we can convert vectors to _n_-by-1 matrices, which is often a worthwhile way of thinking about them.

* We can replace our `into_matrix` methods for matrix slices with `From`/`Into`. (Although a counterargument against this is that type inference can't always infer what you mean by `my_slice.into()`, whereas `my_slice.into_matrix()` didn't have that problem. Also, if we were going to do this, one might expect us to also similarly replace the `into_vec` methods on matrices and vectors, but that turns out to be subtler than it first appears due to trait coherence rules forbidding implementation of a trait not defined in the current crate for a type not defined in the current crate.)

### DIFF
--- a/rusty-machine/src/learning/nnet.rs
+++ b/rusty-machine/src/learning/nnet.rs
@@ -29,7 +29,7 @@
 //! model.train(&inputs, &targets);
 //!
 //! let test_inputs = Matrix::new(2,3, vec![1.5,1.5,1.5,5.1,5.1,5.1]);
-//! 
+//!
 //! // And predict new output from the test inputs
 //! model.predict(&test_inputs);
 //! ```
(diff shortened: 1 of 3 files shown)


AtheMathmo:
Thanks Zack!

I really like using the convert tools - it works really well for `Vector` as you stated.

Personally I think we should keep the `into_matrix` method in addition to having the conversion tools present. This is because there are other explicit into methods within the std library, e.g. [into_vec](https://doc.rust-lang.org/std/primitive.slice.html#method.into_vec) for slices.

If you strongly disagree I can be convinced though :)

zackmdavis:
(amended)

PR #79, opened 2016-06-19
Fixing diagonal cov option bug. Adding gmm constructor

AtheMathmo:
Fixing bug with diagonal covariance options - the weight factor was not included.

Also added a new constructor that allows the user to specify the prior mixture weights.

### DIFF
--- a/rusty-machine/src/learning/gmm.rs
+++ b/rusty-machine/src/learning/gmm.rs
@@ -132,6 +132,36 @@ impl GaussianMixtureModel {
         }
     }
 
+    /// Constructs a new GMM with the specified prior mixture weights.
+    ///
+    /// The mixture weights must have the same length as the number of components.
+    /// The sum of the mixture weights must be 1.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use rusty_machine::learning::gmm::GaussianMixtureModel;
+    /// use rusty_machine::linalg::vector::Vector;
+    ///
+    /// let mix_weights = Vector::new(vec![0.25, 0.25, 0.5]);
+    ///
+    /// let _ = GaussianMixtureModel::with_weights(3, mix_weights);
+    /// ```
+    pub fn with_weights(k: usize, mixture_weights: Vector<f64>) -> GaussianMixtureModel {
+        assert!(mixture_weights.size() == k, "Mixture weights must have length k.");
+        assert!((mixture_weights.sum() - 1f64).abs() < 1e-12);
+
+        GaussianMixtureModel {
+            comp_count: k,
+            mix_weights: mixture_weights,
+            model_means: None,
+            model_covars: None,
+            log_lik: 0f64,
+            max_iters: 100,
+            cov_option: CovOption::Full,
+        }
+    }
+
     /// The model means
     ///
     /// Returns a reference to the Option<Matrix<f64>>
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Including the weight breaks the GMM example as a matrix can no longer be decomposed. This is likely an issue with the example rather than the model code.

Will investigate and update here.

rusty-machine/src/learning/gmm.rs
```diff
@@ -132,6 +132,36 @@ impl GaussianMixtureModel {
...
+    /// ```
+    /// use rusty_machine::learning::gmm::GaussianMixtureModel;
+    /// use rusty_machine::linalg::vector::Vector;
+    ///
+    /// let mix_weights = Vector::new(vec![0.25, 0.25, 0.5]);
+    ///
+    /// let _ = GaussianMixtureModel::with_weights(3, mix_weights);
+    /// ```
+    pub fn with_weights(k: usize, mixture_weights: Vector<f64>) -> GaussianMixtureModel {
+        assert!(mixture_weights.size() == k, "Mixture weights must have length k.");
```
zackmdavis: Panicking is kind of a big deal; would it be better to return an `Option<GaussianMixtureModel>` or `Result<GaussianMixtureModel, rusty_machine::Error>` with a note that many callers will want to tack on a `.expect("I know what I'm doing")`?

AtheMathmo: I'm not too sure about this.. I don't think an `Option` is appropriate here. Result may be but I tend to consider those to be for recoverable errors. I don't see a situation where a Result would be more helpful than panicking?

Of course - detecting this at compile time would be ideal. But I'm unconvinced that letting the user face the panic when they unpack the result is any better? (If anything it makes the library a tiny bit more annoying to use?)

Please feel free to argue back! I should definitely be using Results in more places...

zackmdavis: >  I don't see a situation where a Result would be more helpful than panicking?

If someone is writing a script that calls this function with a fixed argument, then there is no such situation. But as a library author, you don't know what your callers might want: the guts of some application five years from now could invoke this with `mixture_weights` computed from the result of some other function applied to some other function applied to user input. The terrifying discipline of monadic error handling is precisely that the programmer is forced to explicitly acknowledge through the type system that the entire thread could blow up _right here_ if they're not sure that the argument is correct and they don't supply something else to do instead. It's definitely _not_ the most ergonomic on average. But data-mungers who care about ergonomics more than correctness should really be working in Python.

zackmdavis: ([`PathBuf.to_str`](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.to_str) returning `None` if the path isn't valid Unicode is an example of something in the standard library returning an Option even when the typical case will almost always be `Some`. I agree that in this case we would want a `Result`.)

AtheMathmo: I'm not sure the second case you describe would be especially common. Though from this discussion I've realised it is probably best to normalize the mixture weights rather than assert they sum to 1.

I guess I'm struggling to imagine a situation in which the user will discover the result is not `ok` and actually do something about it. If you try to make a new GMM and it fails I'm not sure what you can do? Perhaps you could dump any intermediate data to file before aborting?

With all that said; you probably are right. However, I haven't respected this elsewhere in the library and I would prefer to do this consistently if we bring it in. I'm considering trying it here to see how it goes though.

> But data-mungers who care about ergonomics more than correctness should really be working in Python.

Though I like python - this did make me chuckle :D.

AtheMathmo:
@zackmdavis Have opened #81 to help keep track of the issues you presented above in addition to the wider problem. I will merge this for now but will do my best to amend it soon (time permitting...)

PR #80, opened 2016-06-28
A couple samples

ic:
This PR contains 2 samples based on the SVM and NNET APIs, plus a `README.md`, as we discussed the thing together.

The PR attempts to address the couple points we discussed in this thread so far.

Note the NNET sample is not the easiest we can do, but I have tried to make it a bit exciting (as in "meaningful") as well as self-contained.


rusty-machine/examples/nnet-and_gate.rs
```diff
@@ -0,0 +1,89 @@
...
+    const THRESHOLD: f64 = 0.7;
+
+    const SAMPLES: usize = 10000;
+    println!("Generating {} training data and labels...", SAMPLES as u32);
+
+    let mut input_data = Vec::with_capacity(SAMPLES * 2);
+    let mut label_data = Vec::with_capacity(SAMPLES);
+
+    for _ in 0..SAMPLES {
+        let Open01(left) = random::<Open01<f64>>();
```
AtheMathmo: I think perhaps some inline comments here could be helpful. If people don't read the README I can see this being confusing. (Just a very brief explanation of what we're doing).

rusty-machine/examples/nnet-and_gate.rs
```diff
@@ -0,0 +1,89 @@
...
+        ];
+    let test_inputs = Matrix::new(test_cases.len() / 2, 2, test_cases);
+    let res = model.predict(&test_inputs);
+
+    println!("Evaluation...");
+    let mut hits = 0;
+    let mut misses = 0;
+    // Evaluation
+    println!("Got\tExpected");
+    //for (idx, prediction) in res.into_vec().iter().enumerate() {
```
AtheMathmo: I think you meant to exclude this chunk?

AtheMathmo:
This is really fantastic stuff, thank you!

I have a couple minor comments on the nnet example (as line-notes). Otherwise this is ideal!

One last thought - is it worth us updating the `.travis.yml` to *build* these examples? I'm genuinely not sure - but it may be good to ensure they always compile? (Emphasis on build as running the examples could lead to very slow CI).

AtheMathmo:
@ic - I just saw that you merged master into the PR (thanks!).

Before merging this PR I was waiting on a response to some of my previous comments. I'm actually pretty happy with the PR as it is if you don't have the time to add more. Let me know if you want to make any more tweaks and if not I'll merge this and start playing with a few more examples/tutorials of my own!

ic:
@AtheMathmo Sorry, I have tried to take some time earlier to improve according to your comments. No time yet :-( But I am keen to continue (well, I can't use Rust at work for now, but I see good reasons to).

AtheMathmo:
@ic - no problem at all. No need to rush - just wanted to check that you were happy to make the changes yourself.

Take your time :)

rusty-machine/examples/nnet-and_gate.rs
ic: On this one, I fixed it, but the original intention was to show two ways to evaluate the network output (we have no softmax and co right now). The problem, as you saw, was that I committed exactly the same code!

As it was just a tiny "plus", I have just removed the commented section.

ic:
As for Travis, is it possible to spawn a separate build just for the examples? This would avoid slowing down the main test suite, as the examples *are* slower.

AtheMathmo:
@ic - I'm not sure about spawning up a new build on travis. I imagine it would be possible by playing around with the [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix).

I think it may be a good idea to get this PR merged without worrying about the CI for now. I can create a ticket to track the travis stuff in the mean time.

Thanks for these examples - it's a really fantastic start!

AtheMathmo:
@ic - Turns out `cargo test` will automatically build the examples directory but not run it. This is pretty ideal. Looks like we wont need to do any additional travis magic!

ic:
Oh, yes! I remember reading that in the docs. This is really ideal.

Issue #81, opened 2016-06-29
Error Handling


AtheMathmo:
Error handling is pretty poor at the moment. Lots of functions can panic which should perhaps be replaced with `Result` and `Option`.

This issue should track some of this work and provide a central spot for reporting instances of _panic abuse_.

**Known issues:**

- [x] Decomposition should return results instead of panicking. See #16 .
- [x] The `with_weights` constructor introduced by #79 . 
- [x] Initialization in K-means can cause invalid states. As described in #60 .

**Other tasks:**

- [x] Create a new `error` module to handle errors in the learning module.
- [x] Create a new `error` module to handle errors in the linalg module.
- [x] Make the `train` functions in `SupModel` and `UnSupModel` return a Result. Training can fail - not panicking would be nice...

AtheMathmo:
I made the decision to keep the `with_weights` constructor panicking. I'll close this out but error handling should still get some attention!

PR #83, opened 2016-06-30
K means and error handling

### LINKED ISSUE #82: K-Means Custom Initialization
We should create a trait to define an initialization algorithm for K-Means. The user could then choose the existing algorithms or create their own.

AtheMathmo:
This PR should resolves #82 and part of #81 .

## Error Handling

It adds a new error handling module which will allow us to utilize results readily through the learning module. I kept this error module to just the learning side of rusty-machine as I plan to pull the linalg part into a new crate soon<sup>tm</sup>.

## K-Means restructuring

It also restructures the K-Means module by providing a trait for the initialization algorithm. This has the benefit of providing more flexibility to the user. The downside is an arguably more complex interface (not convinced of this). And more importantly we lose the Builder model we had before - this is because the type of the initialization algorithm must be confirmed as soon as we create the struct.

## Breaking changes

The above means some breaking changes to K-means. In addition to that described above I have made the fields on the KMeansClassifier private (with helper methods providing access).

If you have time (and the interest) it would be good to get some feedback from @zackmdavis and @danlrobertson  . You were both involved with the K-means discussion in #30 a while ago. Of course feedback from anyone else is welcome!

I'm not sure if this should be merged now and swept under the rug as a 0.3.X release or included with further changes later. I'm leaning towards breaking the rules again...

### DIFF
--- a/rusty-machine/src/learning/k_means.rs
+++ b/rusty-machine/src/learning/k_means.rs
@@ -187,7 +187,7 @@ impl UnSupModel<Matrix<f64>, Vector<usize>> for KMeansClassifier {
 
     /// Train the classifier using input data.
     fn train(&mut self, inputs: &Matrix<f64>) {
-        self.init_centroids(inputs);
+        self.init_centroids(inputs).unwrap();
         let mut cost = 0.0;
         let eps = 1e-14;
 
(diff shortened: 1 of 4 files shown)


rusty-machine/src/learning/k_means.rs
```diff
@@ -278,13 +205,11 @@ impl KMeansClassifier {
...
-
-            for j in classes.data().iter() {
-                if *j == i {
-                    vec_i.push(*j);
-                }
-            }
+            let vec_i: Vec<usize> = classes.data()
+                                           .iter()
+                                           .filter(|&x| *x == i)
+                                           .map(|x| *x)
```
AtheMathmo: should use `.cloned()` here.

rusty-machine/src/learning/k_means.rs
```diff
@@ -366,28 +297,25 @@ impl KMeansClassifier {
...
-            let mut vec_i = Vec::new();
-
-            for j in &random_assignments {
-                if *j == i {
-                    vec_i.push(*j);
-                }
-            }
+            let vec_i: Vec<usize> = random_assignments.iter()
+                                                      .filter(|&x| *x == i)
+                                                      .map(|x| *x)
```
AtheMathmo: Should use `.cloned()` here.

Issue #85, opened 2016-07-01
Use Axes enum for Matrix Mean and Variance


AtheMathmo:
Currently we use a `usize` input to determine which axis to take the mean/variance over. Instead we should use the `Axes` enum in the `matrix` module.

Relevant code: [`Axes`](https://github.com/AtheMathmo/rusty-machine/blob/v0.3.1/rusty-machine/src/linalg/matrix/mod.rs#L22-L29), [`mean`](https://github.com/AtheMathmo/rusty-machine/blob/v0.3.1/rusty-machine/src/linalg/matrix/mod.rs#L863) and [`variance`](https://github.com/AtheMathmo/rusty-machine/blob/v0.3.1/rusty-machine/src/linalg/matrix/mod.rs#L902).

This should be an easy change. Most the work will be chasing up existing calls in the library and updating them.

*Bonus*:

There is probably some optimization that can be done in these methods. Feel free to tackle that... e.g. use `set_len` on the vector `t` in `variance`, and then set by index (unchecked). This is safe as `T: Float (:Copy)`. This is pretty minor though and can be ignored for this ticket and rolled out into a new ticket.

DarkDrek:
Hi I want to this simple change.

AtheMathmo:
Awesome!  I checked the commit and it looks good to me. Can you open up a PR - if the CI passes I'll be more than happy to merge it.

Thanks!

DarkDrek:
I didn't know that examples are also tested in the first version of the commit. So I did another one...

AtheMathmo:
The PR looks good and so I've merged. Thanks!

PR #88, opened 2016-07-03
Use Use Axes enum for Matrix Mean and Variance #85

DarkDrek:


AtheMathmo:
@DarkDrek - Thanks! This is great. I'll merge the PR now.

I'm happy to release this as part of a 0.3.X release as the breakage is very minor and shallow.

PR #89, opened 2016-07-03
Use unsafe set_len and direct indexing in variance

DarkDrek:
This was suggested in #85 

I'm not sure if it was worth it to pull out the `t[j] `outside of the `match` what is your opinion?

### DIFF
--- a/rusty-machine/src/linalg/matrix/mod.rs
+++ b/rusty-machine/src/linalg/matrix/mod.rs
@@ -916,10 +916,12 @@ impl<T: Float + FromPrimitive> Matrix<T> {
             let mut t = Vec::<T>::with_capacity(m);
 
             unsafe {
+                t.set_len(m);
+                
                 for j in 0..m {
-                    match axis {
-                        Axes::Row => t.push(*self.data.get_unchecked(i * m + j)),
-                        Axes::Col => t.push(*self.data.get_unchecked(j * n + i)),
+                    t[j] = match axis {
+                        Axes::Row => *self.data.get_unchecked(i * m + j),
+                        Axes::Col => *self.data.get_unchecked(j * n + i),
                     }
 
                 }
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Thanks for another PR! This one has some minor caveats though.

I believe this may still have bound checks on the t[j] indexing - which I believe will prevent the compiler vectorizing. To get around it you can either do an unsafe `get_unchecked_mut(j)` and then mutate this value. Or you can use a trick like [this one used elsewhere](https://github.com/AtheMathmo/rusty-machine/blob/master/rusty-machine/src/linalg/utils.rs#L143-L156).

The best way to check whether this is working as expected is to pull up some minimal version of the code in the [rust playground](https://play.rust-lang.org/) and check the output assembly to see if the code was vectorized and bound checks removed. Pretty messy stuff :) I'll take a go at this myself to see how it looks and report back.

As for your question about moving the t[j] out - I think it looks more Rust-like so I'm all for it!

AtheMathmo:
**TLDR; Looks like the code you presented actually optimizes well!**

Here is a [minimal example](https://play.rust-lang.org/?gist=ecd42463a111eb625a005adaff298462&version=nightly&backtrace=0) in the playground. Here I've replaced your assignment with a new slice (`t_slice`). If you go to the MIR view (**make sure you are compiling in Release not Debug**) you can see that there is a bounds check (search for "index out of bounds").

However if we go back to [your code](https://play.rust-lang.org/?gist=738669cc9e58d894cacb30673040eb1e&version=nightly&backtrace=0). The bounds check is no longer there and the code vectorizes well (looking at the asm output there is a lot of movup [mov packed scalar] and use of xmm registry).

**As one last point...** Could you change the line `let mut variance = Vector::new(vec![T::zero(); m]);` to `let mut variance = Vector::zeros(m);`. That one is my bad code :) but I think this makes it a little more readable.

DarkDrek:
I changed that line. Praise the compiler ;) the simple code won. Maybe we can also simplify [that](https://github.com/AtheMathmo/rusty-machine/blob/master/rusty-machine/src/linalg/utils.rs#L143-L156)?

AtheMathmo:
I'll definitely look at it. When I wrote that code the rejoicing was definitely needed - perhaps compiler changes means that it no longer helps..

Thanks for the PR. Merging now.

PR #95, opened 2016-07-05
Use Option<&Type> instead of &Option<Type>

### LINKED ISSUE #92: Linear and Logistic Regression clone parameters
Currently the methods to return the parameters look like this:

```rust
pub fn parameters(&self) -> Option<Vector<f64>> {
    match *self.base.parameters() {
        Some(ref p) => Some(p.clone()),
        None => None,
    }
}
```

It would be more appropriate to return an `&Option<Vector<f64>>`. This would allow the user to control when to clone the data. The new methods will look like this:

```rust
pub fn parameters(&self) -> &Option<Vector<f64>> {
    &self.parameters
}
```

DarkDrek:
More methods provided by Option can be used this way
Fixes #92 

Should the documentation `Returns a reference to the Option<...>` also be changed?

### DIFF
--- a/rusty-machine/src/learning/gmm.rs
+++ b/rusty-machine/src/learning/gmm.rs
@@ -183,8 +183,8 @@ impl GaussianMixtureModel {
     /// Returns a reference to the Option<Matrix<f64>>
     /// which contains the model means. Each row represents
     /// the mean of one of the Gaussians.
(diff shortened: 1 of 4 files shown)


AtheMathmo:
Looks like you got them all.

I think that the documentation should be changed. Perhaps to:

`Returns an Option<&...> containing the model parameters`

Where `parameters` can be changed to whatever the variable is.

Issue #94, opened 2016-07-05
Implement `IntoIterator` for `MatrixSlice` and `MatrixSliceMut`


AtheMathmo:
We currently implement an `iter()` method for each of the above. However it would be nice to be able to consume the slice into an iterator so that we release ownership of the underlying data.

This issue should use the [`IntoIterator`](https://doc.rust-lang.org/std/iter/trait.IntoIterator.html) trait. The existing slice `iter()` methods should suffice within these implementations.

DarkDrek:
I'm a bit confused `MatrixSlice` doesn't seem to have ownership of data it just borrows data from a `Matrix`?

AtheMathmo:
Yes, that's correct.

I'm actually a little unsure about this issue - shortly after writing it I had doubts. My initial thoughts were:

- Consuming the slice into an iterator may be useful.
- Being able to use the `for _ in slice` syntax would be nice.
- In general `IntoIterator` is a commonly used and valuable trait to implement.

But in hindsight I'm not sure that the first is particularly useful. And additionally the second is a little confusing - would every user expect the iteration over a matrix to be over elements in row-major order?

I'd like to hear your thoughts on it too.

DarkDrek:
- Consuming the slice doesn't work since it implements `Copy`
- Since it is a view for collection I would expect to be able to iterate over it naturally. The (default) iteration order has to be documented somewhere. The simple syntax wouldn't be good when we want to equally support different ways of iterating (row or column based).

Overall I'm still in favor of supporting it since row-major order will be the default because it's faster. So if we have column-major iteration the user will have to explicitly request it.

AtheMathmo:
Yeah you are right with the first point - though `MatrixSliceMut` does not implement copy. Perhaps it should...

My point about iteration was more about whether users would expect to iterate over rows, or single elements? For example - iterating over a numpy array (`for x in arr:`) will iterate over rows. Of course we don't need to follow this convention - but I wanted to raise the question anyway.

DarkDrek:
Since we have other methods to iterate over rows (iter_rows) I find i more intuitive that `for x in arr` gives me single elements and the other ways of iterating are explicit. I would expect that `arr.into_iter()` uses roughly the same logic as `arr.iter()`

AtheMathmo:
Yes I think I agree. In which case - I think it should be enough to simple call `arr.iter()` inside the `IntoIterator` implementation.

PR #99, opened 2016-07-07
Implement IntoIterator for MatrixSlice

DarkDrek:
Fixes part of #94 but first I want to discuss some things with this implementation.

1. I changed `MatrixSlice.iter()` to take an external lifetime bound. It seems to compile this way but I'm not sure if it is correct. This was necessary since the deducted lifetime was invalid `into_iter` takes `self` as `Self` and so the borrowed value was only alive inside the method.
2. I duplicated the implementation for all three cases `value`, `&` and `&mut` this could surly be solved with some macro-magic. Would be nice if someone could help me there.
3. I'm not sure how to test this. The only thing that comes in mind is that I could manually iterate and check each value individually but that seems like a lot of duplicated code.

Small compile test
```rust
fn into_test() { 
    let a = Matrix::new(3, 3, vec![2.0; 9]); 
    let mut b = MatrixSlice::from_matrix(&a, [1, 1], 2, 2);
    
    for v in b { 
    } 
    
    {
        let x = &b; 
        for v in x { 
        }
    }
    
    {
        let y = &mut b;
        for v in y { 
        }
    }
} 
```

### DIFF
--- a/rusty-machine/src/linalg/matrix/iter.rs
+++ b/rusty-machine/src/linalg/matrix/iter.rs
@@ -3,6 +3,7 @@ use std::marker::PhantomData;
 use std::slice;
 
 use super::{Matrix, MatrixSlice, MatrixSliceMut};
+use super::slice::{SliceIter, SliceIterMut};
 
 /// Row iterator.
 #[derive(Debug)]
@@ -322,6 +323,60 @@ impl<'a, T: 'a + Copy> FromIterator<&'a [T]> for Matrix<T> {
     }
 }
 
+impl<'a, T> IntoIterator for MatrixSlice<'a, T> {
+    type Item = &'a T;
+    type IntoIter = SliceIter<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+impl<'a, T> IntoIterator for &'a MatrixSlice<'a, T> {
+    type Item = &'a T;
+    type IntoIter = SliceIter<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+impl<'a, T> IntoIterator for &'a mut MatrixSlice<'a, T> {
+    type Item = &'a T;
+    type IntoIter = SliceIter<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+impl<'a, T> IntoIterator for MatrixSliceMut<'a, T> {
+    type Item = &'a mut T;
+    type IntoIter = SliceIterMut<'a, T>;
+
+    fn into_iter(mut self) -> Self::IntoIter {
+        self.iter_mut()
+    }
+}
+
+impl<'a, T> IntoIterator for &'a MatrixSliceMut<'a, T> {
+    type Item = &'a T;
+    type IntoIter = SliceIter<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter()
+    }
+}
+
+impl<'a, T> IntoIterator for &'a mut MatrixSliceMut<'a, T> {
+    type Item = &'a mut T;
+    type IntoIter = SliceIterMut<'a, T>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.iter_mut()
+    }
+}
+
 #[cfg(test)]
 mod tests {
 
@@ -459,4 +514,46 @@ mod tests {
 
     }
 
+    #[test]
+    #[allow(unused_variables)]
+    fn into_iter_compile() { 
+        let a = Matrix::new(3, 3, vec![2.0; 9]); 
+        let mut b = MatrixSlice::from_matrix(&a, [1, 1], 2, 2);
+    
+        for v in b { 
+        } 
+    
+        for v in &b { 
+        } 
+    
+        for v in &mut b { 
+        } 
+    } 
+    
+    #[test]
+    #[allow(unused_variables)]
+    fn into_iter_mut_compile() { 
+        let mut a = Matrix::new(3, 3, vec![2.0; 9]); 
+        
+        {
+            let b = MatrixSliceMut::from_matrix(&mut a, [1, 1], 2, 2);
+    
+            for v in b { 
+            } 
+        }
+    
+        {
+            let b = MatrixSliceMut::from_matrix(&mut a, [1, 1], 2, 2);
+    
+            for v in &b { 
+            } 
+        }
+    
+        {
+            let mut b = MatrixSliceMut::from_matrix(&mut a, [1, 1], 2, 2);
+    
+            for v in &mut b { 
+            } 
+        }
+    } 
 }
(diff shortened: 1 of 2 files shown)


AtheMathmo:
Thanks for another PR!

Perhaps this requires a little more thinking than I had first anticipated! Responding to each of your points:

1. I'll have to dig around a little more to check - but I'm not sure this should be necessary. For example look at the [implementation](https://github.com/rust-lang/rust/blob/4b89debc7b776a947331dd0ceaa52ba8476157b2/src/libcore/slice.rs#L203) for [slices](https://github.com/rust-lang/rust/blob/4b89debc7b776a947331dd0ceaa52ba8476157b2/src/libcore/slice.rs#L779). They seem to use the [data lifetime](https://github.com/rust-lang/rust/blob/4b89debc7b776a947331dd0ceaa52ba8476157b2/src/libcore/slice.rs#L919) and can call `iter()` without issues. I feel like we want to tie the iterator to the data lifetime or we may face problems - the iterator should not be able to out live the underlying data (imagine a user consuming the matrix inside of the iterator block).

2. I'm not too sure if we actually need all of these? Perhaps we do.. Another thing I'll have to dig into a little . If we do need them then macros may help but it isn't too bad as is.

3. For testing I would just add a method similar to the one you have in the description above. Just so that in future if anything changes the compiler will shout when building the tests. The only existing tests for `iter()` are the doc tests right now - if you think we should add more in the tests module feel free.

As one final note, will you be able to extend this work to `MatrixSliceMut` as well? After we resolve the implementation questions of course!

AtheMathmo:
With regards to **1**:

I tested by removing the `'b` lifetime and added the following test:

```rust
#[test]
fn into_test() { 
    let a = Matrix::new(3, 3, vec![2.0; 9]); 
    let mut b = MatrixSlice::from_matrix(&a, [1, 1], 2, 2);

    for v in b { 
        println!({}, v);
    } 

    for v in &b { 
        println!({}, v);
    } 

    for v in &mut b { 
        println!({}, v);
    } 
} 
```

This compiles for me without issue. Perhaps I am misunderstanding,could you explain again the issue you encountered with the lifetime?

DarkDrek:
1. In your example they implement `IntoIterator` for `& Something` and not for `Something`. If we look at [vec.into_iter()](https://github.com/rust-lang/rust/blob/4b89debc7b776a947331dd0ceaa52ba8476157b2/src/libcollections/vec.rs#L1349) the version without references doesn't call iter but uses a different kind of iterator that with some unsfae magic consumes the `vec`. In our case we can't do that the slice doesn't own the data. It is interesting because if I copy the implementation of `iter()` it seems to work.

```rust
impl<'a, T> IntoIterator for MatrixSlice<'a, T> {
    type Item = &'a T;
    type IntoIter = SliceIter<'a, T>;

    fn into_iter(self) -> Self::IntoIter {
        SliceIter {
            slice_start: self.ptr,
            row_pos: 0,
            col_pos: 0,
            slice_rows: self.rows,
            slice_cols: self.cols,
            row_stride: self.row_stride,
            _marker: PhantomData::<&'a T>,
        }
    }
}
```
PS: Did you ran your test? I get 
```
src/linalg/matrix/slice.rs:224:9: 224:13 error: `self` does not live long enough
src/linalg/matrix/slice.rs:224         self.iter()
                                       ^~~~
src/linalg/matrix/slice.rs:223:42: 225:6 note: reference must be valid for the lifetime 'a as defined on the block at 223:41...
src/linalg/matrix/slice.rs:223     fn into_iter(self) -> Self::IntoIter {
                                                                        ^
src/linalg/matrix/slice.rs:223:42: 225:6 note: ...but borrowed value is only valid for the scope of function body at 223:41
src/linalg/matrix/slice.rs:223     fn into_iter(self) -> Self::IntoIter {
```

All of this is also applicable to `MatrixSliceMut` and if everything is clear I would add a implementation here.

DarkDrek:
2. We should allow the user to use all three forms used in the test. The user may not even know what type the `MatrixSlice` exactly has due to type interference.

AtheMathmo:
1. Well this is interesting... I did indeed run my test and it worked fine. For your convenience here is a [dump of the file](https://gist.github.com/AtheMathmo/e195be35007ac5f63abbbc5f8f6d2a95). Is the difference that I have explicitly used lifetime `'a`? Otherwise maybe a different rustc version (I'm using 1.9.0 stable)?. \[It is an odd rule, you can also just number everything 1 and it autoincrements.\]
1. I agree with you on this one. It isn't harmful to allow iterating over all of them - but iterating explicitly over a mutable slice may look odd.

DarkDrek:
1. Yes that was the key difference with `'a` in the `iter()` return type it also compiles here (rust nightly ... from a few days ago ...).

DarkDrek:
`MatrixSliceMut` has also it's own specialty... If we only have a non-mutable reference to it we can't iterate over it with a mutable reference as result so it returns only a `SliceIter`.

AtheMathmo:
Ah - that makes sense. It does make me question whether the restriction on `iter_mut` to take an `&mut MatrixSliceMut` is necessary though? Is it safe for this method to simple take an `&MatrixSliceMut`?

I imagine there is some case where it is. And if this is the case I am happy to merge this PR :).

The only other thing worth asking is whether we should move this iteration stuff into the `matrix/iter` module. I'm happy with it being kept where it is for now though.

DarkDrek:
For starters it at least compiles without `mut`. For me the decision on this boils down to whether we want a non-mutable borrower of `MatrixSliceMut` to modify the underlying `Matrix`. Maybe the separation of the mutability to different classes is the problem? Combined with Rust's mutability handling we have mutable `MatrixSlice`'s and non-mutable `MatrixSliceMut`'s... Do we really need `MatrixSliceMut` isn't a `&mut MatrixSlice` adequate for that. But maybe I'm missing something? Does this concept exist somewhere else inside the Rust standard library?

I would move it to that file it fits there better.

AtheMathmo:
That would indeed be better - and fit in more with how the `core::slice` code works. I couldn't figure out how to get it to work well though. Without getting into too much detail - I couldn't see how to solve problems like constructing a mutable `MatrixSlice`. I also used [ndarray](https://github.com/bluss/rust-ndarray) as help and bluss handles this in the same way.

Do you want to move the iteration methods to that file now?

DarkDrek:
I can see the problems... but at least this PR will work with what we have now.

Yes I want to do that now as the last part before the merge.

rusty-machine/src/linalg/matrix/iter.rs
```diff
@@ -459,4 +514,46 @@ mod tests {
...
+            let b = MatrixSliceMut::from_matrix(&mut a, [1, 1], 2, 2);
+    
+            for v in &b { 
+            } 
+        }
+    
+        {
+            let mut b = MatrixSliceMut::from_matrix(&mut a, [1, 1], 2, 2);
+    
+            for v in &mut b {
```
AtheMathmo: I think it's worth adding an

`*v = 1.0;`

here. This will add a further check to the compilation that the value returned is mutable. (And also the same for the `for v in b` loop.

AtheMathmo:
In addition to the inline comment, a nitpick: I'd rather have `for _ in b` than allow unused variables. I think this case it makes more sense.

PR #103, opened 2016-07-17
DBSCAN

### LINKED ISSUE #100: Implement DBSCAN Algorithm
This would prove another welcome addition to rusty-machine. Resources:

- [Wikipedia](https://en.wikipedia.org/wiki/DBSCAN)
- [Sklearn](http://scikit-learn.org/stable/modules/generated/sklearn.cluster.DBSCAN.html)

AtheMathmo:
This PR resolves #100 .

The PR adds a simple implementation of the DBSCAN algorithm. There is some hacky work around getting the `predict` method of the `UnSupModel` trait to work. We need to clone the data to do prediction and so we make this heavy task gated behind a `predictive` property of the model which must be explicitly set.

In future it may be a good idea to add a `Clustering` trait which does not require the `predict` method. We can then implement this trait for DBSCAN/K-Means/Nearest neighbours. The only obvious downside is that we'd need universal function syntax to distinguish if we used the same function names (or we use different names which may add some complication).

### DIFF
--- a/src/learning/dbscan.rs
+++ b/src/learning/dbscan.rs
@@ -55,6 +55,8 @@ impl UnSupModel<Matrix<f64>, Vector<usize>> for DBSCAN {
 impl DBSCAN {
     /// Create a new DBSCAN model.
     pub fn new(eps: f64, min_points: usize) -> DBSCAN {
+        assert!(eps > 0f64, "The model epsilon must be positive.");
+
         DBSCAN {
             eps: eps,
             min_points: min_points,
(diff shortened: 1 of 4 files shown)


AtheMathmo:
For now I've added a note to the module documentation stating that this should be treated as experimental for now.

If anybody has any good ideas about how we can adapt the current code I'd be happy to hear it. Otherwise I'll wait until Nearest Neighbours is implemented to get a better feel for the overall requirements.

PR #105, opened 2016-07-23
Nnet rewrite

tafia:
Small rewrite of nnet module.

- in `create_weights`: use `rand::distributions::range::Range` as [advised on rand doc](https://doc.rust-lang.org/rand/rand/trait.Rng.html#method.gen_range)
  - removed `initialize_weights` function and put the logic directly in `create_weights`
- in `compute_grad`: remove an  unecessary intermediary `grad` vector. On my very unscientific test, running the example `nnet-and_gate` is 2.5 % faster.

### DIFF
--- a/src/learning/nnet.rs
+++ b/src/learning/nnet.rs
@@ -315,10 +315,9 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
             }
         }
 
-        let mut grad = Vec::with_capacity(self.layer_sizes.len() - 1);
-        let mut capacity = 0;
+        let mut gradients = Vec::new();
 
-        for (l, activ_item) in activations.iter().enumerate().take(self.layer_sizes.len() - 1) {
+        for (l, activ_item) in activations[..self.layer_sizes.len() - 1].iter().enumerate() {
             // Compute the gradient
             let mut g = deltas[self.layer_sizes.len() - 2 - l].transpose() * activ_item;
 
@@ -330,14 +329,7 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
                 g += zeros.vcat(&self.criterion.reg_cost_grad(non_bias_weights));
             }
 
-            capacity += g.cols() * g.rows();
-            grad.push(g / (inputs.rows() as f64));
-        }
-
-        let mut gradients = Vec::with_capacity(capacity);
-
-        for g in grad {
-            gradients.append(&mut g.into_vec());
+            gradients.append(&mut (g / inputs.rows() as f64).into_vec());
         }
 
         // Compute the cost
@@ -345,9 +337,9 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
 
         // Add the regularized cost
         if self.criterion.is_regularized() {
-            for i in 0..self.layer_sizes.len() - 1 {
-                cost += self.criterion.reg_cost(self.get_non_bias_weights(weights, i));
-            }
+            cost += (0..self.layer_sizes.len() - 1)
+                .map(|i| self.criterion.reg_cost(self.get_non_bias_weights(weights, i)))
+                .sum();
         }
 
         (cost, gradients)
--- a/src/learning/nnet.rs
+++ b/src/learning/nnet.rs
@@ -53,7 +53,8 @@ use learning::toolkit::regularization::Regularization;
 use learning::optim::{Optimizable, OptimAlgorithm};
 use learning::optim::grad_desc::StochasticGD;
 
-use rand::{Rng, thread_rng};
+use rand::thread_rng;
+use rand::distributions::{Sample, range};
 
 /// Neural Network Model
 ///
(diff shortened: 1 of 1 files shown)


AtheMathmo:
Thanks for the PR! It will take a me a little while to figure out all the changes but I'm happy to hear we get some performance gains from this!

src/learning/nnet.rs
```diff
@@ -343,24 +325,17 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
...
         let mut cost = self.criterion.cost(&activations[activations.len() - 1], targets);
 
         // Add the regularized cost
         if self.criterion.is_regularized() {
-            for i in 0..self.layer_sizes.len() - 1 {
-                cost += self.criterion.reg_cost(self.get_non_bias_weights(weights, i));
-            }
+            cost += (0..self.layer_sizes.len() - 1)
+                .map(|i| self.criterion.reg_cost(self.get_non_bias_weights(weights, i)))
+                .sum();
```
AtheMathmo: The `sum` method is unstable which is causing the travis build to fail. If using an iterator like this gives significant gains in performance then you can use `fold` instead? Though as the function within the loop is fairly intense I'm not sure it would make much difference.

src/learning/nnet.rs
```diff
@@ -328,10 +311,9 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
             }
         }
 
-        let mut grad = Vec::with_capacity(self.layer_sizes.len() - 1);
-        let mut capacity = 0;
+        let mut gradients = Vec::new();
 
-        for (l, activ_item) in activations.iter().enumerate().take(self.layer_sizes.len() - 1) {
+        for (l, activ_item) in activations[..self.layer_sizes.len() - 1].iter().enumerate() {
```
AtheMathmo: Maybe it is just me but personally I prefer `take` to indexing like this. The compiler probably boils them down to the same thing but I worry indexing may add some extra bound checks or something?

(Feel free to make the case for indexing instead!)

src/learning/nnet.rs
```diff
@@ -328,10 +311,9 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
             }
         }
 
-        let mut grad = Vec::with_capacity(self.layer_sizes.len() - 1);
-        let mut capacity = 0;
+        let mut gradients = Vec::new();
```
AtheMathmo: I can't figure out very quickly but we should be able to compute the capacity and initialize `with_capacity`.

src/learning/nnet.rs
```diff
@@ -292,8 +275,8 @@ impl<'a, T: Criterion> BaseNeuralNet<'a, T> {
 
                 a = ones.hcat(&a);
 
-                activations.push(a.clone());
-                z = a * self.get_layer_weights(weights, l);
+                z = &a * self.get_layer_weights(weights, l);
+                activations.push(a);
```
AtheMathmo: Nice spot!

AtheMathmo:
I've given a few comments - I may have a couple more things as I digest more.

I've also got to commend you on working through the code in the nnet module! It's pretty messy and difficult to follow - thanks so much for putting the time in.

src/learning/nnet.rs
tafia: It doesn't make a difference. It felt more like rust way to me. 
I'll revert back

AtheMathmo: I agree - the `sum` method would be more rust-like. Hopefully it lands in stable soon!

src/learning/nnet.rs
tafia: I'm not an expert of llvm but I cannot spot a difference between the 2. I will revert back to `take` if you prefer.

AtheMathmo: It's fine to leave it as is. Unless [clippy](https://github.com/Manishearth/rust-clippy) complains I'm happy to keep the indexing :)

src/learning/nnet.rs
tafia: I'm new in machine learning, but I'd say we'll have as many `gradients` as we have input `weights` ?

tafia:
Pushed a new commit with your comments. I set `gradients` capacity to `weights.len()`. I *think* it is correct.

AtheMathmo:
@tafia I'm not sure if that is correct as the each layer's weights will contain some bias weights which will always be equal to 1 (and so not updated by the gradient).

It's too late in the night for me to figure out the maths, but I imagine it will be something like `weights.len() - self.layer_sizes.iter().take(self.layer_sizes.len() - 1).sum()` - where sadly we cannot use sum :). That is because the left column of each layer's weights will be bias weights - which means all but the last layer will contribute it's size in bias weights.

I'll have to actually verify that... Also it isn't critical - we can always just `shrink_to_fit` after the fact.

Thanks again for your hard work on this.

> I'm new in machine learning...

You've jumped into some pretty rudimental stuff here - hopefully it wasn't too painful :).

tafia:
if we look at the [regularization part only](https://github.com/AtheMathmo/rusty-machine/blob/master/src/learning/nnet.rs#L341-L343), it looks like the bias are included in the gradient matrix, at least this is how I interpreted the first zero row. 

This is confirmed when I print `gradient.len()` and `weights.len()`.

AtheMathmo:
Yes you are absolutely correct! I believe this wraps up the PR then, thanks once more.

tafia:
Great!
Thanks, I will continue learning on the code. It is very interesting!

AtheMathmo:
Just wanted to add that if there's anything I can do to help you figure out the code please don't hesitate to reach out!

tafia:
Thanks!!
Much appreciated

PR #107, opened 2016-07-25
Use more `iter_rows` and `select_rows`

### LINKED ISSUE #93: Replacing row iteration by chunks with `iter_rows` method
In multiple places we use `chunks` iteration where we could use the new `iter_rows` and `iter_rows_mut` methods.

Known places where this exists:

- `gmm.rs` in the `update_params` method (on `new_means.mut_data()`).
- `naive_bayes.rs` in the `update_params` method (on `targets.data()`).
- `naive_bayes.rs` in the `get_classes` method (on `log_probs.data()`).

In addition to the above there are likely a number of other areas where code can be tidied up using the `iter_rows` methods. One such example:

In `svm.rs`, the `ker_mat` method. Instead of using:

```rust
for i in 0..dim` {
    for j in 0..dim2 {
        ker_data.push(self.ker.kernel(&m1.data()[i * cols..(i + 1) * cols],
                                          &m2.data()[j * cols..(j + 1) * cols]));
    }
}
```

We can iterate over the rows in `m1` and `m2` and then push the data into `ker_data`.

**There are likely many more examples. The above is a good start.**

tafia:
Partly closes #93 

I greped `data()` to search for relevant lines. 

Replace with `iter_rows` and `select_rows` whenever I could.

Also sometimes use `zip` iterator instead of `enumerate`. Hopefully it'll lead to less bound checks.

### DIFF
--- a/src/learning/dbscan.rs
+++ b/src/learning/dbscan.rs
@@ -178,11 +178,8 @@ impl DBSCAN {
             let visited = self._visited[*data_point_idx];
             if !visited {
                 self._visited[*data_point_idx] = true;
-                let sub_neighbours =
-                    self.region_query(&inputs.data()[data_point_idx * inputs.cols()..(data_point_idx +
-                                                                                   1) *
-                                                                                  inputs.cols()],
-                                      inputs);
+                let sub_neighbours = self.region_query(inputs.select_rows(&[*data_point_idx]).data(),
+                                                       inputs);
 
                 if sub_neighbours.len() >= self.min_points {
                     self.expand_cluster(inputs, *data_point_idx, sub_neighbours, cluster);
--- a/src/learning/gp.rs
+++ b/src/learning/gp.rs
@@ -122,19 +122,14 @@ impl<T: Kernel, U: MeanFunc> GaussianProcess<T, U> {
     /// Construct a kernel matrix
     fn ker_mat(&self, m1: &Matrix<f64>, m2: &Matrix<f64>) -> Matrix<f64> {
         assert_eq!(m1.cols(), m2.cols());
-        let cols = m1.cols();
 
         let dim1 = m1.rows();
         let dim2 = m2.rows();
 
         let mut ker_data = Vec::with_capacity(dim1 * dim2);
-
-        for i in 0..dim1 {
-            for j in 0..dim2 {
-                ker_data.push(self.ker.kernel(&m1.data()[i * cols..(i + 1) * cols],
-                                              &m2.data()[j * cols..(j + 1) * cols]));
-            }
-        }
+        ker_data.extend(
+            m1.iter_rows().flat_map(|row1| m2.iter_rows()
+                                    .map(move |row2| self.ker.kernel(row1, row2))));
 
         Matrix::new(dim1, dim2, ker_data)
     }
--- a/src/learning/svm.rs
+++ b/src/learning/svm.rs
@@ -103,19 +103,15 @@ impl<K: Kernel> SVM<K> {
     /// Construct a kernel matrix
     fn ker_mat(&self, m1: &Matrix<f64>, m2: &Matrix<f64>) -> Matrix<f64> {
         assert_eq!(m1.cols(), m2.cols());
-        let cols = m1.cols();
 
         let dim1 = m1.rows();
         let dim2 = m2.rows();
 
         let mut ker_data = Vec::with_capacity(dim1 * dim2);
 
-        for i in 0..dim1 {
-            for j in 0..dim2 {
-                ker_data.push(self.ker.kernel(&m1.data()[i * cols..(i + 1) * cols],
-                                              &m2.data()[j * cols..(j + 1) * cols]));
-            }
-        }
+        ker_data.extend(
+            m1.iter_rows().flat_map(|row1| m2.iter_rows()
+                                    .map(move |row2| self.ker.kernel(row1, row2))));
 
         Matrix::new(dim1, dim2, ker_data)
     }
--- a/src/learning/svm.rs
+++ b/src/learning/svm.rs
@@ -150,13 +150,10 @@ impl<K: Kernel> SupModel<Matrix<f64>, Vector<f64>> for SVM<K> {
 
         for t in 0..self.optim_iters {
             let i = rng.gen_range(0, n);
-            let mut sum = 0f64;
-            for j in 0..n {
-                sum += alpha[j] * targets[j] *
-                       self.ker.kernel(&full_inputs.data()[i * m..(i + 1) * m],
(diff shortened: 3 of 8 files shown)


AtheMathmo:
This looks like it adds a lot of really nice changes! Thank you! I need to properly review so these comments may not be valid, but there are a couple caveats with `zip` and `select_rows`.

From memory `zip` can prevent vectorization in some cases - and optimizes funnily with floats. I can't find any references right now but hopefully it isn't important for this PR.

And `select_rows` will `copy` the underlying data. In some places it may be more efficient to index, or slice, or something else instead. I did have plans a while ago to rename this to `copy_rows` or something but never followed through...

I'll review properly tomorrow morning but I hope the above doesn't invalidate anything you've done!

tafia:
For the `zip`, I believe the issue for vectorization has been [fixed](https://github.com/rust-lang/rust/pull/33090) for slices (which is our case), though it may not be in rust stable yet ...
I didn't know about the issue regarding the `float`s.

I didn't notice about the `select_rows` ... this is indeed problematic (even if as you stated it could sometimes lead to faster code). 

I couldn't find a `get_row` fn in rulinalg crate, (a fn that just borrows a slice of the matrix). I think the code is easier to read (and probably to generalized, if ever needed) with such a function than without using the `&matrix.data()[i*col .. (i + 1) * col]`

If needed I think we could as a start create an Trait that adds it to `Matrix`, `MatrixSlice` and `MatrixSliceMut`.

To conclude I think having some benchmarks would probably help taking a decision ...

AtheMathmo:
That zip change is awesome! I believe I remember reading the `float` stuff in a conversation that bluss was involved in - perhaps that has been addressed since too. In this case I'm happy to use `zip` liberally where relevant.

There is an [open issue on rulinalg](https://github.com/AtheMathmo/rulinalg/issues/7) to add `get_rows`. If you think it would help this PR a great deal I'm happy to implement this soon (or of course you can have a go too!).

And yes I agree - benchmarks would help a great deal. Sadly I have no benchmarks for rusty-machine yet - only for rulinalg.

With all that said, I still haven't reviewed this PR. I'll do so soon. Thanks again!

src/learning/dbscan.rs
```diff
@@ -178,11 +178,8 @@ impl DBSCAN {
             let visited = self._visited[*data_point_idx];
             if !visited {
                 self._visited[*data_point_idx] = true;
-                let sub_neighbours =
-                    self.region_query(&inputs.data()[data_point_idx * inputs.cols()..(data_point_idx +
-                                                                                   1) *
-                                                                                  inputs.cols()],
-                                      inputs);
+                let sub_neighbours = self.region_query(inputs.select_rows(&[*data_point_idx]).data(),
```
AtheMathmo: I'm inclined to keep this as is until https://github.com/AtheMathmo/rulinalg/issues/7 has been tackled. I'll add a comment on that ticket to note that this piece of code should be revisited once that feature lands.

Happy to hear your thoughts on it.

src/learning/gp.rs
```diff
@@ -122,19 +122,14 @@ impl<T: Kernel, U: MeanFunc> GaussianProcess<T, U> {
...
 
         let mut ker_data = Vec::with_capacity(dim1 * dim2);
-
-        for i in 0..dim1 {
-            for j in 0..dim2 {
-                ker_data.push(self.ker.kernel(&m1.data()[i * cols..(i + 1) * cols],
-                                              &m2.data()[j * cols..(j + 1) * cols]));
-            }
-        }
+        ker_data.extend(
```
AtheMathmo: This is neat!

I wonder whether this is more efficient than nested loops - either way I doubt it is significant.

src/learning/svm.rs
```diff
@@ -150,17 +146,13 @@ impl<K: Kernel> SupModel<Matrix<f64>, Vector<f64>> for SVM<K> {
...
             let i = rng.gen_range(0, n);
-            let mut sum = 0f64;
-            for j in 0..n {
-                sum += alpha[j] * targets[j] *
-                       self.ker.kernel(&full_inputs.data()[i * m..(i + 1) * m],
-                                       &full_inputs.data()[j * m..(j + 1) * m]);
-            }
-            sum *= targets[i] / (self.lambda * (t as f64));
+            let row_i = full_inputs.select_rows(&[i]);
+            let sum =  full_inputs.iter_rows()
```
AtheMathmo: Previously I had left the `*= targets / (self.lambda * (t as f64))` line out of the loop to reduce rounding errors and number of operations per loop.

I think this is still a sensible thing to do?

src/learning/toolkit/regularization.rs
```diff
@@ -141,11 +141,8 @@ mod tests {
 
         assert!((a - (input_mat.norm() / 12f64)) < 1e-18);
 
-        let true_grad = input_mat.data()
-            .iter()
-            .map(|x| x / 6f64)
-            .collect::<Vec<_>>();
-        for eps in (b - Matrix::new(3, 4, true_grad)).into_vec() {
+        let true_grad = &input_mat / 6f64;
+        for eps in (b - true_grad).into_vec() {
```
AtheMathmo: Thanks for this - not sure what I was thinking with the previous code!

AtheMathmo:
Having thought a little more about the `select_rows` introductions - I think it is best to use the `select_rows` method as in this PR.

Even if it is less efficient - it will be minimally so. And it will be easier to track these down (in addition to some other existing cases I imagine) when the `get_rows` methods do land.

tafia:
Made a PR on rulinalg for `get_row`s:
https://github.com/AtheMathmo/rulinalg/pull/8

I'll update this PR if it gets merged.

src/learning/svm.rs
tafia: I don't get your point.
Here too the `* targets / (self.lambda * (t as f64))` is done *after* the `fold`, so it should be strictly equivalent.

AtheMathmo:
I added a few dictator-esque remarks. Nothing too heavy just personal taste and tests. The code itself looks good and will make this PR even more valuable. Thanks for all your help!

src/learning/svm.rs
AtheMathmo: You are correct. Sorry - my mistake!

tafia:
> I added a few dictator-esque remarks

I think you're right! Always better to follow the same coding style throughout the whole crate, it helps readability.

tafia:
> I'll update this PR if it gets merged.

I actually think we should merge this PR as-is ... and open another issue to replace `select_rows` with `get_rows`

I've done a quick grep, and it looks like it is used widely enough to have its own PR.

AtheMathmo:
Yep, I agree with you too! Will merge this in the next hour or so. Thanks.

Issue #109, opened 2016-07-26
Use more `Borrow`s


tafia:
There are many places where copies are done (via `select_rows` for instance) when a simple borrow would have been enough.

We could probably save some memory by:
- implementing `Borrow` and `AsRef` between `Matrix` and `MatrixSlice(Mut)` (they make sense I think)
- having more generic traits that require `Borrow<T>` or `AsRef<T>` instead of `&T`

Example with `Optimizable`:
```rust
        /// Trait for models which can be gradient-optimized.
        pub trait Optimizable {
            /// The input data type to the model.
            type Inputs;
            /// The target data type to the model.
            type Targets;
            /// Compute the gradient for the model.
            fn compute_grad(&self,
                            params: &[f64],
                            inputs: Borrow<Self::Inputs>,
                            targets: Borrow<Self::Targets>)
                            -> (f64, Vec<f64>);
        }
```

Then we could use `MatrixSlice` instead of `select_rows` [here](https://github.com/AtheMathmo/rusty-machine/blob/master/src/learning/optim/grad_desc.rs#L181-L183).

There are probably many examples.

AtheMathmo:
This is a really, really good suggestion. It will remove a lot of unneeded copying - especially in conjunction with your `get_row` PR that will be merged soon.

I have yet to use `Borrow` and `AsRef` so the differences are lost on me right now. I'll have to do some reading.

I'm happy to work on this issue if you didn't want to.

tafia:
Please do it! You may come up with new suggestions

AtheMathmo:
Ok I'll tackle this one :). I won't be able to tackle it until the weekend (focusing on a presentation for a meetup).

AtheMathmo:
Added a [mirror issue on rulinalg](https://github.com/AtheMathmo/rulinalg/issues/14) to track the implementation there.

AtheMathmo:
Closing this out. We've gone with an alternate approach and wont be able to do something like this until custom DSTs land.

PR #111, opened 2016-07-27
Remove several select_rows

### LINKED ISSUE #110: Use rulinalg's `get_row` methods instead of `select_rows` on a single row
Currently `select_row` copies data whereas `get_row` provides a slice of the row.

We should switch to using `get_row` where possible. Additionally we should use the unchecked methods as often as we safely can.

tafia:
Partly closes #110 

Made some minor rewriting, I can roll them back if needed.
Also changed one signature on `Distribution` trait
```rust
fn update_params(&mut self, data: &Matrix<f64>, class: usize)
```
The fn now borrows `data` instead of taking ownership (there were no functional reason I think).

For the remaining `select_rows` one pattern that appears a lot is repeating one particular row and then doing matix operations with it. It looked like a good thing to do so I didn't touch it (I don't know how fast matrix operation is compared to a regular for loop).

### DIFF
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -293,7 +293,7 @@ impl GaussianMixtureModel {
             for i in 0..n {
                 let inputs_i = MatrixSlice::from_matrix(inputs, [i, 0], 1, d);
                 let diff = inputs_i - new_means_k; 
-                cov_mat += self.compute_cov(&diff, membership_weights[[i, k]]);
+                cov_mat += self.compute_cov(diff, membership_weights[[i, k]]);
             }
             new_covs.push(cov_mat / sum_weights[k]);
 
@@ -303,7 +303,7 @@ impl GaussianMixtureModel {
         self.model_covars = Some(new_covs);
     }
 
-    fn compute_cov(&self, diff: &Matrix<f64>, weight: f64) -> Matrix<f64> {
+    fn compute_cov(&self, diff: Matrix<f64>, weight: f64) -> Matrix<f64> {
         match self.cov_option {
             CovOption::Full => (diff.transpose() * diff) * weight,
             CovOption::Regularized(eps) => (diff.transpose() * diff) * weight + eps,
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -31,8 +31,7 @@
 //! println!("{:?}", post_probs.data());
 //! ```
 
-use linalg::Vector;
-use linalg::Matrix;
+use linalg::{Matrix, MatrixSlice, Vector};
 use rulinalg::utils;
 
 use learning::UnSupModel;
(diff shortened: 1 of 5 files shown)


src/learning/gmm.rs
```diff
@@ -302,7 +303,7 @@ impl GaussianMixtureModel {
         self.model_covars = Some(new_covs);
     }
 
-    fn compute_cov(&self, diff: Matrix<f64>, weight: f64) -> Matrix<f64> {
+    fn compute_cov(&self, diff: &Matrix<f64>, weight: f64) -> Matrix<f64> {
```
AtheMathmo: It doesn't make a difference in this case but it may in the future. For example we _could_ do in-place operations for transpose or `elemul`. It's not happening right now though so no reason to change this.

src/learning/k_means.rs
```diff
@@ -209,14 +209,13 @@ impl<InitAlg: Initializer> KMeansClassifier<InitAlg> {
...
-        for i in 0..self.k {
-            let vec_i: Vec<usize> = classes.data()
-                .iter()
-                .enumerate()
-                .filter(|&(_, &x)| x == i)
-                .map(|(idx, _)| idx)
-                .collect();
-            
+
+        let mut row_indexes = vec![Vec::new(); self.k];
```
AtheMathmo: Could you explain the motivation for this change? I'm not sure which is better but if you have some insight I'd love to find out!

AtheMathmo:
Thanks for the PR!

This looks good. But is there any reason why you did not include the grad_desc module? If it requires special treatment I'll write up a new issue.

src/learning/k_means.rs
tafia: Current code has `O(self.k * classes.data().len())` complexity. New one has `O(classes.data().len())`.

tafia:
> But is there any reason why you did not include the grad_desc module

The only reason is that you need a `Matrix` as per the `Optimizable` trait. You cannot use a `MatrixSlice`.
Linked to #109

src/learning/k_means.rs
tafia: By the way, there is [some code](https://github.com/AtheMathmo/rusty-machine/blob/master/src/learning/k_means.rs#L309-L312) I didn't understand.
To me it is missing a `enumerate()`, like the part we're commenting. Else we could rewrite a simpler
```rust 
let mut random_assignments = vec![1; k];
for _ in k..inputs.rows() {
    random_assignments[rng.gen_range(0, k)] += 1;
}
// ...
let vec_i = vec![i; random_assignments[i]];
```

AtheMathmo: Yes you're right - your code is more efficient.

As for the code you don't understand, that is because it is a bug! I fixed the same issue elsewhere in a recent release. I'd welcome a fix in this PR as you've made a similar improvement in the code this thread belongs to.

AtheMathmo:
We have the `get_row` method for `Matrix` too though? My apologies if I'm missing something. I was planning on addressing this with #109 too - but I think we can get a small improvement now too?

src/learning/k_means.rs
tafia: Fine I'll do it

AtheMathmo: Thanks for cleaning up my mess :) (and also for spotting it)

tafia:
You mean changing the signature of `compute_grad`?

AtheMathmo:
Sorry - it took me a while but I _finally_ understand the problem...

I'll address this with #109 . I'll also stop chucking more things for you to onto this PR and merge it once the last standing item is in. Thanks for your patience.

src/learning/k_means.rs
tafia: It's ok, don't worry! I've just pushed a fix (please double check it works as expected)

AtheMathmo:
The bug fix looks good, thanks! Merging now.

PR #116, opened 2016-08-04
Adding data transformers for preprocessing

### LINKED ISSUE #112: Implement some basic data transformations
It would be nice to begin supporting some basic data transformations. Things like scaling, etc. A model similar to sickit learn's transform interface would be a good jumping off point.

The following transformations would be nice for a start:

- Centering around zero with unit variance.
- Scaling Min/Max to some range.

To achieve this we should have a new trait defining our transform API. Something like this:

```rust
pub trait Transformer<T> {
    /// Transforms the inputs and stores the transformation in the Transformer
    fn transform(&mut self, inputs: T) -> Result<T, learning::Error>;

    /// Reverses the transform to recover the original data
    fn inv_transform(&self, inputs: T) -> Result<T, learning::Error>;
}
```

So our min max scaler may look like:

```rust
pub struct MinMaxScaler {
    /// Mins per column of input data
    input_min: Vec<usize>,
    /// Maxs per column of input data
    input_max: Vec<usize>,
    /// The min of the new data (default 0)
    scaled_min: f64,
    /// The max of the new data (default 1)
    scaled_max: f64,
}

impl Transformer<Matrix<f64>> for MinMaxScaler {
    fn transform(&mut self, inputs: Matrix<f64>) -> Result<Matrix<f64>, learning::Error> {
        // Find min and max of each column of features
        // For each column:
        //    Subtract the respective input_min and add the scaled_min
        //    Set each value x to x * (scaled_max / input_max)
    }

    fn inv_transform(&self, inputs: Matrix<f64>) -> Result<Matrix<f64>, learning::Error> {
        // Should fail if the Transformer has not been used to transform data yet
        //
        // Do the inverse of the transform:
        // For each column:
        //    Set each value x to x * (input_max / scaled_max)
        //    Subtract the scaled_min and add input_min
    }
}
```

The above is just to illustrate how this may look. Further requirements may mean something slightly different is needed. Some caveats we should be aware of:

- How to handle constant features? (If the std dev is zero we don't want to divide by this in scaling)
- How to handle class data? (i.e. values 0,1,2,3 representing one of 4 classes)

There are probably some other cases we should consider. But I think for now we can capture these as fail cases or just ignore them completely...

AtheMathmo:
This is a work in progress but I'm opening for feedback now. This will eventually resolve #112 .

In order to complete this PR I need to do the following:

- Implement a scaler for zero centering and unit variance.
- Add error handling in min/max case when column is a constant feature. I think for now just fail gracefully..?
- Add better documentation and review module layout.

### DIFF
--- a/src/data/transforms/minmax.rs
+++ b/src/data/transforms/minmax.rs
@@ -144,6 +144,23 @@ mod tests {
     }
 
     #[test]
+    fn custom_scale_test() {
+        let inputs = Matrix::new(2, 2, vec![-1.0f32, 2.0, 0.0, 3.0]);
+
+        let mut scaler = MinMaxScaler::new(1.0, 3.0);
+        let transformed = scaler.transform(inputs).unwrap();
+
+        assert!(transformed.data().iter().all(|&x| x >= 1.0));
+        assert!(transformed.data().iter().all(|&x| x <= 3.0));
+
+        // First column scales to zero and second to 1
+        (transformed[[0, 0]] - 1.0).abs() < 1e-10;
+        (transformed[[0, 1]] - 1.0).abs() < 1e-10;
+        (transformed[[1, 0]] - 3.0).abs() < 1e-10;
+        (transformed[[1, 1]] - 3.0).abs() < 1e-10;
+    }
+
+    #[test]
     fn constant_feature_test() {
         let inputs = Matrix::new(2, 2, vec![1.0, 2.0, 1.0, 3.0]);
 
--- a/src/data/transforms/minmax.rs
+++ b/src/data/transforms/minmax.rs
@@ -1,0 +1,124 @@
+//! The Min-Max transformer
+
+use learning::error::{Error, ErrorKind};
+use linalg::Matrix;
+use super::Transformer;
+use std::f64;
+
+/// The min max scaler
+#[derive(Debug)]
+pub struct MinMaxScaler {
+    /// Mins per column of input data
+    input_min: Vec<f64>,
+    /// Maxs per column of input data
+    input_max: Vec<f64>,
+    /// The min of the new data (default 0)
+    scaled_min: f64,
+    /// The max of the new data (default 1)
+    scaled_max: f64,
+}
+
+impl MinMaxScaler {
+    /// Constructs a new MinMaxScaler with the specified scale.
+    pub fn new(min: f64, max: f64) -> MinMaxScaler {
+        MinMaxScaler {
+            input_min: Vec::new(),
+            input_max: Vec::new(),
+            scaled_min: min,
+            scaled_max: max,
+        }
+    }
+}
+
+impl Transformer<Matrix<f64>> for MinMaxScaler {
+    fn transform(&mut self, mut inputs: Matrix<f64>) -> Result<Matrix<f64>, Error> {
+        let features = inputs.cols();
+
+        self.input_min = vec![f64::MAX; features];
+        self.input_max = vec![f64::MIN; features];
+
+        for row in inputs.iter_rows() {
+            for (idx, feature) in row.into_iter().enumerate() {
+                if !feature.is_finite() {
+                    return Err(Error::new(ErrorKind::InvalidData,
+                                          format!("Data point {0} in column {1} cannot be \
+                                                   processed",
+                                                  feature,
+                                                  idx)));
+                }
+
+                if *feature < self.input_min[idx] {
+                    self.input_min[idx] = *feature;
+                }
+
+                if *feature > self.input_max[idx] {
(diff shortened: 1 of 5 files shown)


src/data/transforms/minmax.rs
```diff
@@ -0,0 +1,247 @@
...
+/// which allows us to transform the input data to have a new minimum
+/// and maximum per column.
+///
+/// See the module description for more information.
+#[derive(Debug)]
+pub struct MinMaxScaler<T: Float> {
+    /// Mins per column of input data
+    scale_factors: Option<Vec<T>>,
+    /// Maxs per column of input data
+    const_factors: Option<Vec<T>>,
```
tafia: Why not aligning comments and names (`min_factors` and `max_factors`) ?

AtheMathmo: Thanks for reading this over.

This is an artifact of me changing the name of the field. I'll update now, thanks! (These comments wont be visible in docs as the fields aren't public - but still well worth getting right).

AtheMathmo:
@tafia - does the rest of this PR look ok to you? If so I'll merge.

It doesn't feel quite right to me but is a good first draft for people to try out.

tafia:
I am busy at the moment. I'll have another look tomorrow.
Feel free to merge I'll open an issue if needed.

AtheMathmo:
No problem - I'll wait.

I'm away this weekend so wont be very responsive myself.

AtheMathmo:
@tafia any other comments on this? If not I'll merge (if you don't have time yet but do want to take another look through that's fine - I can wait)

tafia:
Sorry forgot it.
Looking now

src/data/transforms/minmax.rs
```diff
@@ -0,0 +1,247 @@
...
+    scaled_min: T,
+    /// The max of the new data (default 1)
+    scaled_max: T,
+}
+
+/// Create a default `MinMaxScaler` with minimum of `0` and
+/// maximum of `1`.
+impl<T: Float> Default for MinMaxScaler<T> {
+    fn default() -> MinMaxScaler<T> {
+        MinMaxScaler {
```
tafia: I would prefer using `MinMaxScaler::new` here. 
It is simpler if the `new` function ever changes ...

AtheMathmo: I'm not sure I understand. There is a `new` function below this one - this is an implementation of the `Default` trait. I think it makes sense to have this here as I think scaling between 0 and 1 is common enough to warrant it as a default.

src/data/transforms/minmax.rs
```diff
@@ -0,0 +1,247 @@
...
+                if !feature.is_finite() {
+                    return Err(Error::new(ErrorKind::InvalidData,
+                                          format!("Data point in column {} cannot be \
+                                                   processed",
+                                                  idx)));
+                }
+
+                if *feature < input_min[idx] {
+                    input_min[idx] = *feature;
+                }
```
tafia: Very small point: 
You can avoid bound checks and `enumerate` using `zip` ... and only one zip if you work with tuples.
```rust
let mut input_min_max = vec![(T::min_value(), T::masx_value()); features];
```
... and computing the `scales` is simpler as well:
```rust
let scales = try!(input_min_max.iter().map(|(x, y)| {
    let s = (self.scaled_max - self.scaled_min) / (y - x);
    if s.is_finite() { 
        Ok(s) 
    } else {  
        Err(Error::new(ErrorKind::InvalidData,"Constant feature columns not supported."))
    }
}).collect::<Result<Vec<_>, _>>());
```

src/data/transforms/minmax.rs
```diff
@@ -0,0 +1,247 @@
...
+        }
+
+        let consts = input_max.iter()
+            .zip(scales.iter())
+            .map(|(&x, &s)| self.scaled_max - x * s)
+            .collect::<Vec<_>>();
+
+        for row in inputs.iter_rows_mut() {
+            for i in 0..features {
+                row[i] = scales[i] * row[i] + consts[i];
```
tafia: `utils::in_place_vec_bin_op` ?

src/data/transforms/minmax.rs
tafia:
```rust
impl<T: Float> Default for MinMaxScaler<T> {
    fn default() -> MinMaxScaler<T> {
        MinMaxScaler::new(T::zero(), T::one())
    }
}
```

AtheMathmo: Ah I see - thanks!

AtheMathmo:
I pushed some updates from your feedback. Would you mind taking another look?

PR #118, opened 2016-08-29
Add benches from examples

tafia:
Copied the examples and bench train/predict.
At least this is a start so we have something to compare to ...

On my system, it gives:
```
test bench_svm_sign_learner ... bench:      32,314 ns/iter (+/- 1,221)
test bench_nnet_and_gate ... bench:  31,395,040 ns/iter (+/- 1,461,709)
test bench_k_means ... bench:  12,792,701 ns/iter (+/- 4,426,994)
```

### DIFF
--- a/benches/k-means_generating_cluster.rs
+++ b/benches/k-means_generating_cluster.rs
@@ -1,0 +1,65 @@
+#![feature(test)]
+
+extern crate rusty_machine;
+extern crate rand;
+extern crate test;
+
+use rusty_machine::linalg::Matrix;
+use rusty_machine::learning::k_means::KMeansClassifier;
+use rusty_machine::learning::UnSupModel;
+
+use rand::thread_rng;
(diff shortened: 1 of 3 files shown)


AtheMathmo:
This is a great start. I have some minor suggestions.

I'd prefer to see this in a sub module, I.e. `benches/examples/*.rs`.

And also i think we should split out benchmarks for training and predicting where possible. So for k-means train and check centroid shin one benchmark and then predict new classes in another.

What do you think?

tafia:
Fine. I'll do this tomorrow

AtheMathmo:
@tafia do you want me to take this one over? Don't want to rush you but I'd like to get the rulinalg update in so I can add a few other breaking changes I'd been holding out on :)

tafia:
Hi
Sorry I probably won't be able to do it in the coming days, i'm not at
home. Go ahead if you can.

PR #121, opened 2016-09-03
Implemented RMSProp Algorithm

NivenT:
I added [RMSProp](http://sebastianruder.com/optimizing-gradient-descent/index.html#rmsprop) as another gradient descent algorithm. I think I have heard that it has become a popular optimization algorithm recently, so I thought it would nice to add. I tested it on the simple model in the tests folder, and on the neural network example.

### DIFF
--- a/src/learning/optim/grad_desc.rs
+++ b/src/learning/optim/grad_desc.rs
@@ -391,14 +391,16 @@ impl<M> OptimAlgorithm<M> for RMSProp
                                                       &inputs.select_rows(&[*i]),
                                                       &targets.select_rows(&[*i]));
 
-                let grad = Vector::new(grad);
+                let mut grad = Vector::new(grad);
                 let grad_squared = grad.clone().apply(&|x| x*x);
                 // Update cached average of squared gradients
                 rmsprop_cache = &rmsprop_cache*self.decay_rate + &grad_squared*(1.0 - self.decay_rate);
                 // RMSProp update rule 
-                params = &params - (&grad*self.learning_rate).elediv(
-                                        &rmsprop_cache.clone().apply(&|x| (x + self.epsilon).sqrt()));
-                
+                utils::in_place_vec_bin_op(grad.mut_data(), rmsprop_cache.data(), |x, &y| {
+                    *x = *x * self.learning_rate / (y + self.epsilon).sqrt();
+                });
+                params = &params - &grad;
+
                 end_cost += cost;
             }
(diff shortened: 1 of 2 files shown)


AtheMathmo:
This is really cool! Thanks for the thorough implementation.

I had a quick check through and it looks great. I'll take a more thorough look tomorrow and hopefully get this merged.

src/learning/optim/grad_desc.rs
```diff
@@ -303,10 +303,121 @@ impl<M: Optimizable<Inputs = Matrix<f64>, Targets = Matrix<f64>>> OptimAlgorithm
...
+                                                      &inputs.select_rows(&[*i]),
+                                                      &targets.select_rows(&[*i]));
+
+                let grad = Vector::new(grad);
+                let grad_squared = grad.clone().apply(&|x| x*x);
+                // Update cached average of squared gradients
+                rmsprop_cache = &rmsprop_cache*self.decay_rate + &grad_squared*(1.0 - self.decay_rate);
+                // RMSProp update rule 
+                params = &params - (&grad*self.learning_rate).elediv(
+                                        &rmsprop_cache.clone().apply(&|x| (x + self.epsilon).sqrt()));
```
AtheMathmo: You can avoid the clone here by using the `utils::vec_bin_op` from `rulinalg`. Something like:

`params = &params - utils::in_place_vec_bin_op((&grad * self.learning_rate).mut_data(), rmsprop_cache.data(), |x, &y| {*x =  *x * (y + self.epsilon).sqrt() })`

It's verbose but more efficient. Essentially you scale `grad` and then mutate this data without any cloning etc. It will also be vectorized etc.

AtheMathmo:
I have one minor comment to help prevent some allocation - otherwise this looks good!

NivenT:
Thanks for the suggestion. That clone was bothering me but I was not sure how to get rid of it.

AtheMathmo:
I guess the rulinalg API needs a little refining so that it is more obvious how to achieve things like that...

NivenT:
Maybe. To be completely honest, I did not spend a lot of time looking for a way to do it. How does it look now?

AtheMathmo:
Looks good to me. I'm just going to try out the algorithm in the neural net example to double check that nothing odd happens. If it runs well I'll be happy to merge!

AtheMathmo:
It looks good - thanks for this! RMSProp looks really cool :)

PR #130, opened 2016-09-20
don't hide `stats` module behind feature flag?

zackmdavis:
This got in my way (first was puzzled by the compiler insisting it doesn't exist despite it being documented, then later wasn't aware and still am not aware of how to specify a feature with `cargo run --example`). Unless there's a specific rationale (saving compile time?), it may be more convenient to just make it a first-class part of library.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,7 +207,6 @@ pub mod learning {
     }
 }
 
-#[cfg(feature = "stats")]
 /// Module for computational statistics
 pub mod stats {
 


AtheMathmo:
I'll try to explain why this is feature-gated though it may not be a good justification!

Primarily it is because the stats module is very lacking and I knew I didn't have time to fully build it out. I wanted some basic functionality there but really it barely provides more than the `rand` crate. I was planning on adding more but it looks like other people have already built out some [better solutions](https://github.com/boxtown/statrs).

I haven't really decided what the future of the stats module should be and so I've just let it sit as I focus on other stuff. I'm fairly sure I don't want it to be included by default yet - as right now I feel more inclined to remove it.

To use the feature you build with:

`cargo build --features stats`

(As always feel free to disagree with me and let me know why!)

zackmdavis:
OK

> `--features stats`

(This works with `cargo run --example`, which is what I needed for #131.)

Issue #128, opened 2016-09-19
naïve Bayes classifier example


zackmdavis:
I want to learn about this anyway, so this is to register my intent to put together an example

AtheMathmo:
Awesome! Thank you very much :)

In case you'd like some extra resources I have a [blog post I wrote](http://athemathmo.github.io/2016/04/08/naive-bayes-rusty-machine.html) about the rusty-machine implementation. Otherwise feel free to ask any questions or criticisms of the code here.

PR #131, opened 2016-09-21
example: classifying dogs with a naïve Bayes model

zackmdavis:
This in the matter of #128.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -9,6 +9,7 @@ This directory gathers fully-fledged programs, each using a piece of
 * [K-Means](#k-means)
 * [SVM](#svm)
 * [Neural Networks](#neural-networks)
+* [Naïve Bayes](#naïve-bayes)
 
 ## The Examples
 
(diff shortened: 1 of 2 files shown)


AtheMathmo:
I love the context for the example! I have a few comments for the example code. I think it needs some more inline comments and hand holding. Maybe we can try to separate out the data-creation and the machine learning parts too. But really it only needs some mild changes.

I guess the bigger issue is the feature flag... I'll see if I can find a way to fix that using the `stats` feature (maybe we can hide the example behind the feature flag?). Otherwise it might be best to just use the `rand` crate here for the Gaussians?

zackmdavis:
OK; sometime in the next few days I'll add comments and change it to use [`rand::distributions::normal::Normal`](https://doc.rust-lang.org/rand/rand/distributions/normal/struct.Normal.html)

AtheMathmo:
I've merged in #133 which has some breaking changes (going into the 0.5.0 release). Let me know if you need any help resolving. It's mostly modifications to trait imports and some behind-the-scenes naive bayes changes (shouldn't affect you).

zackmdavis:
(updated and force-pushed)

examples/naive_bayes_dogs.rs
```diff
@@ -0,0 +1,148 @@
...
+        // add a row to the target matrix indicating color
+        let color_row = match training_dog.color {
+            Color::Red => Matrix::new(1, 2, vec![1., 0.]),
+            Color::White => Matrix::new(1, 2, vec![0., 1.]),
+        };
+        target_matrix = target_matrix.vcat(&color_row);
+    }
+
+    // Train!
+    let mut model = NaiveBayes::<naive_bayes::Gaussian>::new();
```
AtheMathmo: I think we should give some more information to the reader here (briefly of course). Why do we choose `Gaussian`?

examples/README.md
```diff
@@ -123,3 +124,27 @@ Got  Expected
 Hits: 4, Misses: 0
 Accuracy: 100%
 ```
+
+### Naïve Bayes
+
+#### Dog Classification
+
+Suppose we have a population composed of red dogs and white dogs, whose friendliness, furriness, and speed can be measured. The group of white dogs is friendlier, furrier, and slower than red dogs by one standard deviation (respectively), but given the color of a dog, friendliness, furriness, and speed are independent of each other. We can use a naïve Bayes model to try to predict dog color given friendliness, furriness, and speed as observations.
```
AtheMathmo: I think another paragraph with a little more technical detail would be good. In particular I think it's worthwhile to address:

- What kind of Naive Bayes' do we use? Why?
- As the example is quite large - maybe walk through roughly what we do? Randomly generate some dogs, convert them into `Matrix` form for rusty-machine. Create model, train model, predict from model on unknown dogs. Assess accuracy.

What do you think? Perhaps there is already enough information there for new users - it certainly makes sense to me but I'm not a good reference point!

zackmdavis:
OK; more commentary later

AtheMathmo:
@zackmdavis did you need anything from me for this?

Not rushing you at all - take as long as you like!

zackmdavis:
Alternatively, merge now and accept improvements later if I (or—we can unrealistically hope—someone else) happen to get around to it??

PR #133, opened 2016-09-21
Updating to Rulinalg 0.3.2

AtheMathmo:
This will resolve PR #117 .

This continues the great work by @tafia by taking into account new breaking changes from both rusty-machine ( #120 ) and rulinalg (variance returns result).

The changes are mostly minor - modifying imports. There are some structural changes to naive bayes to help error propagation. There is also a performance regression on the neural nets but this has been discussed in #117 . The short of it is:

- `apply` is less efficient but there is an issue on rulinalg.
- `hcat` is less efficient on narrow matrices (as in the example) but more efficient on wide ones.

We are also making modifications to the neural nets with #126 which will provide big performance gains.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,4 +16,4 @@ stats = []
 [dependencies]
 num = { version = "0.1.34", default-features = false }
 rand = "0.3.14"
-rulinalg = "0.3.0"
+rulinalg = "0.3.2"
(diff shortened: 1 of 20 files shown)


tafia:
Great!

AtheMathmo:
Sorry that I had to merge this without you! We can start taking advantage of this new work now though!

PR #136, opened 2016-09-22
Use select_rows in cross_validation

theotherphil:
This improves a minor performance issue (not using extend_from_slice) but leaves the major issues unaddressed - allocating fresh memory for every fold, and copying rows for every fold. I think we decided that copying rows was unavoidable, as most models weren't going to support taking a non-contiguous set of Matrix rows as input. I could avoid the per-fold allocations by allocating a mutable buffer up front and reusing it for each fold, but even this is awkward as it would get in the way of adding parallelism - we can't have multiple threads trying to copy into the same mutable buffer.

I guess the best approach is just to add some benchmarks and profile. Then we can at least check that performance doesn't regress when we add support for more kinds of cross-validation.

### DIFF
--- a/benches/examples/cross_validation.rs
+++ b/benches/examples/cross_validation.rs
@@ -1,0 +1,68 @@
+use rusty_machine::linalg::{Matrix, BaseMatrix};
+use rusty_machine::learning::{LearningResult, SupModel};
+use rusty_machine::analysis::score::row_accuracy;
+use rusty_machine::analysis::cross_validation::k_fold_validate;
+use rand::{thread_rng, Rng};
+use test::{Bencher, black_box};
+
+fn generate_data(rows: usize, cols: usize) -> Matrix<f64> {
+    let mut rng = thread_rng();
+    let mut data = Vec::with_capacity(rows * cols);
+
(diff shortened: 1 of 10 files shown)


theotherphil:
Also, I guess I need to learn git properly - this probably shouldn't be showing as including 37 commits!

theotherphil:
I've also fixed some compiler warnings in benchmarks. In the laziest way possible...

AtheMathmo:
> Also, I guess I need to learn git properly - this probably shouldn't be showing as including 37 commits!

It's no problem - I'll squash them before merging. I think the issue arises because you're working off of your old branch and I squashed the changes going in. You'd need to rebase on master I guess?

Thanks for fixing the compiler warning - I wonder if we should wrap the `model.train` calls in `black_box`? If we're not using the output this might get optimized away? (I know this isn't really your problem to fix, sorry!)

theotherphil:
Sure, I'll add the wrapping

theotherphil:
I've committed the compiler warnings fixes in a separate pull request. I think this one should be good for merging now.

AtheMathmo:
Great, thank you!

PR #141, opened 2016-09-23
Add cross-validation benchmarks

theotherphil:
Results on my machine (2.5 GHz Intel Core i7, 16 GB 1600 MHz DDR3):

```
running 4 tests
test examples::cross_validation::bench_1000_10_10  ... bench:     365,266 ns/iter (+/- 30,927)
test examples::cross_validation::bench_1000_10_100 ... bench:   3,708,769 ns/iter (+/- 1,097,906)
test examples::cross_validation::bench_1000_10_3   ... bench:     144,208 ns/iter (+/- 35,749)
test examples::cross_validation::bench_10_10_3     ... bench:       2,036 ns/iter (+/- 415)
```

### DIFF
--- a/benches/examples/cross_validation.rs
+++ b/benches/examples/cross_validation.rs
@@ -1,0 +1,68 @@
+use rusty_machine::linalg::{Matrix, BaseMatrix};
+use rusty_machine::learning::{LearningResult, SupModel};
+use rusty_machine::analysis::score::row_accuracy;
+use rusty_machine::analysis::cross_validation::k_fold_validate;
+use rand::{thread_rng, Rng};
+use test::{Bencher, black_box};
+
+fn generate_data(rows: usize, cols: usize) -> Matrix<f64> {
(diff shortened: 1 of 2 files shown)


AtheMathmo:
The `benches/examples` directory was meant to be a mirror of the `examples` directory - in that it benchmarks the same models and data. However I think that maybe this benchmark is ok to leave there too?

If you would prefer to move it elsewhere then let me know - otherwise I'm happy to merge.

AtheMathmo:
Oh and I forgot - can you add the benchmarks on your system to the PR description? Mostly for my curiosity :)

theotherphil:
Ah yes, I did wonder why all the benchmarks were nested inside the examples directory. I'm happy to leave it where it is for now. We can trivially reshuffle stuff later when there are more benchmarks.

I've added my results to description.

AtheMathmo:
Perfect, thanks!

PR #142, opened 2016-09-23
Fix #119: Calculate sample covariance for GMM

### LINKED ISSUE #119: GMM should compute the sample covariance of the data for its initial state
Currently we simply use an identity matrix as the initial covariance but in practice this is sub optimal.

andrewcsmith:
I was running into some problems, thought perhaps fixing this issue would fix my problems. Didn't work, but here's a fix for #119 anyway.

### DIFF
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -70,17 +70,25 @@ pub struct GaussianMixtureModel {
 impl UnSupModel<Matrix<f64>, Matrix<f64>> for GaussianMixtureModel {
     /// Train the model using inputs.
     fn train(&mut self, inputs: &Matrix<f64>) -> LearningResult<()> {
+        use rulinalg::matrix::Axes;
+
         // Initialization:
         let k = self.comp_count;
 
-        // TODO: Compute sample covariance of the data.
-        // https://en.wikipedia.org/wiki/Sample_mean_and_covariance#Sample_covariance
-        let mut cov_vec = Vec::with_capacity(k);
-        for _ in 0..k {
-            cov_vec.push(Matrix::identity(inputs.cols()));
+        let means = inputs.mean(Axes::Col);
+
+        let mut cov_mat = Matrix::zeros(inputs.cols(), inputs.cols());
+        let reg_value = 1f64 / (inputs.rows() - 1) as f64;
+
+        for j in 0..inputs.cols() {
+            for k in 0..inputs.cols() {
+                cov_mat[[j, k]] = (0..inputs.rows()).map(|i| {
+                    (inputs[[i, j]] - means[j]) * (inputs[[i, k]] - means[k])
+                }).sum::<f64>() * reg_value;
+            }
         }
 
-        self.model_covars = Some(cov_vec);
+        self.model_covars = Some(vec![cov_mat; k]);
 
         let random_rows: Vec<usize> =
             rand_utils::reservoir_sample(&(0..inputs.rows()).collect::<Vec<usize>>(), k);
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -31,7 +31,7 @@
 //! println!("{:?}", post_probs.data());
 //! ```
 
-use linalg::{Matrix, MatrixSlice, Vector, BaseMatrix, BaseMatrixMut};
+use linalg::{Matrix, MatrixSlice, Vector, BaseMatrix, BaseMatrixMut, Axes};
 use rulinalg::utils;
 
 use learning::{LearningResult, UnSupModel};
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -70,18 +70,17 @@ pub struct GaussianMixtureModel {
 impl UnSupModel<Matrix<f64>, Matrix<f64>> for GaussianMixtureModel {
     /// Train the model using inputs.
     fn train(&mut self, inputs: &Matrix<f64>) -> LearningResult<()> {
-        // Initialization:
-        let k = self.comp_count;
-
-        let means = inputs.mean(Axes::Col);
-
-        let mut cov_mat = Matrix::zeros(inputs.cols(), inputs.cols());
         let reg_value = if inputs.rows() > 1 {
             1f64 / (inputs.rows() - 1) as f64
         } else {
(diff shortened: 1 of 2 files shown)


src/learning/gmm.rs
```diff
@@ -70,17 +70,25 @@ pub struct GaussianMixtureModel {
 impl UnSupModel<Matrix<f64>, Matrix<f64>> for GaussianMixtureModel {
     /// Train the model using inputs.
     fn train(&mut self, inputs: &Matrix<f64>) -> LearningResult<()> {
+        use rulinalg::matrix::Axes;
```
AtheMathmo: Can you move this to the `use matrix::{...}` at the top, please.

src/learning/gmm.rs
```diff
@@ -70,17 +70,25 @@ pub struct GaussianMixtureModel {
...
 
-        // TODO: Compute sample covariance of the data.
-        // https://en.wikipedia.org/wiki/Sample_mean_and_covariance#Sample_covariance
-        let mut cov_vec = Vec::with_capacity(k);
-        for _ in 0..k {
-            cov_vec.push(Matrix::identity(inputs.cols()));
+        let means = inputs.mean(Axes::Col);
+
+        let mut cov_mat = Matrix::zeros(inputs.cols(), inputs.cols());
+        let reg_value = 1f64 / (inputs.rows() - 1) as f64;
```
AtheMathmo: We will need to check here that we have `inputs.rows() > 1` and return an `Error` otherwise.

We may well do this check later when we call `variance` but we should catch it here before panicking on underflow (or producing `NaN` cov values when dividing by `0`).

src/learning/gmm.rs
```diff
@@ -70,17 +70,25 @@ pub struct GaussianMixtureModel {
...
+
+        for j in 0..inputs.cols() {
+            for k in 0..inputs.cols() {
+                cov_mat[[j, k]] = (0..inputs.rows()).map(|i| {
```
AtheMathmo: We can _safely_ use `get_unchecked(_mut)` here and below as we know we are within the dimensions.

Though I wonder if it would be better to construct the matrix by pushing to a `Vec` instead of zero-initialization (we access the data sequentially anyway).

src/learning/gmm.rs
```diff
@@ -70,17 +70,25 @@ pub struct GaussianMixtureModel {
...
+                    (inputs[[i, j]] - means[j]) * (inputs[[i, k]] - means[k])
+                }).sum::<f64>() * reg_value;
```
AtheMathmo: I would be inclined to move this `reg_value` multiplication until the very end. It probably makes **very** little difference but I feel as though it may be better to let the compiler vectorize the multiplication.

AtheMathmo:
The code looks good to me. I just need to _sniff test_ it to make sure this isn't leading to any funky results from the GMM. Should have time today to do that.

Thanks!

AtheMathmo:
I just realised there _is_ an issue here - due to the way we handle the `CovOptions` right now.

Essentially the structure is only enforce when we update the covariance. This means that if we start with a full covariance matrix we will try to inverse it once.

andrewcsmith:
Okay, I did some sanity checks to this one, mostly comparing things to numpy's output for covariance of a matrix with itself.

Numpy's [variance method](http://docs.scipy.org/doc/numpy/reference/generated/numpy.var.html#numpy.var) does not give the same result by default, but that's because you need to pass ddof=1 to ensure that it's regularized by `1 / (m.rows()-1)` rather than just `1 / m.rows()`. Other than that the two methods agree.

AtheMathmo:
Thanks for looking into that. I'm feeling good about the changes!

I still haven't had time to run through an actual example and check that nothing funky has gone on but I'll do my best to get that done in the next day or so.

AtheMathmo:
I ran this on some simulated data and it looks good!

I noticed that you have changed the direction over which the mean was computed. I'm a little confused about the axis that we are computing the covariance over - but it seems you've verified that this is correct. As one last safety measure could you compare with [sklearn](https://github.com/scikit-learn/scikit-learn/blob/51a765a/sklearn/mixture/gmm.py#L499)? They use numpy's cov to compute the covariance of `X.T`. I think this is equivalent (+/- difference in divisors) to what we are doing but I'd like to be sure!

Thanks.

andrewcsmith:
Right right, we're computing the mean of all rows, and return a row. That's what `Axes::Row` does correct?

The scikitlearn bit you linked to just computes the covariance and adds the regularization factor to the diagonal. Any idea why you would only add the regularization `min_covar` to the diagonal?

AtheMathmo:
> Right right, we're computing the mean of all rows, and return a row. That's what Axes::Row does correct?

Yep that's right!

> Any idea why you would only add the regularization min_covar to the diagonal?

I'm probably a little lacking on the details but essentially the regularization makes the computations more numerically stable. There is a probabilistic equivalent to that argument about keeping the variance under control too - but I'm not sure if it applies here or exactly what it is :)

We cover the same bases as sklearn but default to full covariance instead of diagonal and don't regularize by default (maybe we should - but then you have to pick appropriate defaults which  becomes tricky).

src/learning/gmm.rs
```diff
@@ -70,17 +70,42 @@ pub struct GaussianMixtureModel {
...
+                    for (k, elem) in row.iter_mut().enumerate() {
+                        *elem = inputs.iter_rows().map(|r| {
+                            (r[j] - means[j]) * (r[k] - means[k])
+                        }).sum::<f64>();
+                    }
+                }
+                cov_mat *= reg_value;
+
+                if let CovOption::Regularized(eps) = self.cov_option {
+                    cov_mat += eps;
```
AtheMathmo: Ah - I just understood your previous question a little more. We should indeed only add this to the diagonal.

For now I guess we can just do this by creating an identity matrix - we'll have [Diagonal iterators](https://github.com/AtheMathmo/rulinalg/pull/55) soon!

andrewcsmith: Gotcha - yes let's fix this later to remove the allocation, but this is good for now.

AtheMathmo:
This looks good to me now. Are you happy to merge?

andrewcsmith:
Yes, I think all the bases are covered here.

PR #149, opened 2016-10-03
`LearningResult` as return type of `membership_weights` in GMM

### LINKED ISSUE #143: GMM panics if the covariance inverse fails
In the `membership_weights` function we compute the inverse of the covariance. This returns a `Result` which we call `expect` on - leading to a panic if the inverse could not be computed.

We should instead make the `membership_weights` function return a `LearningResult` which we `try!` to unwrap in the `train` function. To do this we'll need to map the rulinalg error returned by the `inverse` function to a rusty-machine error.

scholtzan:
I think this resolves #143


AtheMathmo:
This is perfect, thank you!

Issue #138, opened 2016-09-22
Allowing seeds for randomness


AtheMathmo:
We frequently use randomness throughout rusty-machine and in all cases default to the `rand::thread_rng`  (except in the new `Shuffler` added by #135). It is very valuable for users to be able to set a seed to ensure that their models behave the same way on different runs.

It would be nice if we could capture this with minimal effect on the current API. It is probably a requirement that the models (and other components) own their own random number generators. We would likely need to add a new generic type parameter on the models:

```rust
pub struct MLModel<T1, T2, ..., Tn, R> where R: Rng {
    // All the previous fields
   rng: Rng,
}
```

The alternative is to use a Box<Rng> for this field and provide a trait which allows the user to modify the generator for the model:

```rust
pub trait Randomizable {
    fn set_rng(&miut self, rng: Rng);

    fn rng(&self) -> Box<Rng>;
}
```

This approach will keep the API a little cleaner. We can also control access to shared code requiring randomness by using the `Randomizable` trait. For example we could wrap some of the `rand_utils` functions to work on an object which implements `Randomizable`. I'd still like the `rand_utils` functions to be accessed in their current state though.

Thoughts on this are very welcome. I suspect we will need to implement something to get a good idea of the benefits/drawbacks of any approach.

dyule:
Both approaches have the downside of requiring each module that uses it duplicate the code to do with managing the rng.  I can't see any way out of that without some sort of factory for everything that uses randomness, along with lifetimes on the generated objects that would definitely be worse.

Of the two approaches, the second seems much easier for the end user, since it allows for a default of `thread_rng()` if they don't want to bother setting their own.

However, unless I'm misunderstanding box ownership (I might be, it always gets me a bit confused), the second option still results in ownership of the rng.  And of course, it also results in dynamic dispatch, which is slower.

So, I think the choice boils down to either slightly more complexity for the user, or slightly longer runtimes.

I implemented a trial run on LDA, and I've discovered it causes some strange issues with mutability.  That is, any method which uses randomness must be mutable, since all of the methods for generating randomness are mutable.  I'm not sure this makes sense, so I've gone around it using `RefCell`s, but I don't know how great that is either.  Anyways, the relevant pieces are here:

```
pub trait Randomizable {
    fn set_rng(&mut self, rng: Box<Rng>);

    fn rng<'a>(&'a self) -> RefMut<'a, Box<Rng>>;
}

pub struct LDAFitter {
    iterations: usize,
[...]

Issue #122, opened 2016-09-03
Serialisation


theotherphil:
Are there any plans yet for adding support for model serialisation?

AtheMathmo:
I haven't formalized anything yet and I think it will be good to address it after reviewing the API for training (as discussed in #120).

With that said the only real criteria is that it works on stable.

theotherphil:
Ok, thanks.

garro95:
To achieve this, it is better to implement de/serialization on Vector and Matrix from rulinalg before.

PR #151, opened 2016-10-10
Adding an information criterion to the gmm module for dynamic K optimisation

zeryx:
If you want to optimise for K in a cluster model (aka by discovering an optimal fit and to prevent overfitting), you need access to information about the performance of the model.

Right now neither the gaussian mixture model or k-means modules provide direct access to an interface to the fit performance, normally provided as the `log likelihood`.

It definitely is valuable to restrict public access to the interface as much as possible. I don't think the log likelihood itself is a meaningful value and should be encapsulated in the event of internal refactoring, so I implemented the [Bayesian Information Criterion BIC)](https://en.wikipedia.org/wiki/Bayesian_information_criterion) for the gmm module.

One major caveat of the BIC is that for the equation to be meaningful: `n_samples >>> n_clusters`.

If this gets accepted in some form I'll make a second pull request with other information criterions best suited for K optimisation.

### DIFF
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -61,6 +61,7 @@ pub struct GaussianMixtureModel {
     model_means: Option<Matrix<f64>>,
     model_covars: Option<Vec<Matrix<f64>>>,
     log_lik: f64,
+    bic: f64,
     max_iters: usize,
     /// The covariance options for the GMM.
     pub cov_option: CovOption,
@@ -157,6 +158,7 @@ impl GaussianMixtureModel {
             model_means: None,
             model_covars: None,
             log_lik: 0f64,
+            bic: 0f64,
             max_iters: 100,
             cov_option: CovOption::Full,
         }
@@ -199,6 +201,7 @@ impl GaussianMixtureModel {
                 model_means: None,
                 model_covars: None,
                 log_lik: 0f64,
+                bic: 0f64,
                 max_iters: 100,
                 cov_option: CovOption::Full,
             })
@@ -244,6 +247,15 @@ impl GaussianMixtureModel {
     /// ```
     pub fn set_max_iters(&mut self, iters: usize) {
         self.max_iters = iters;
+    }
+
+    /// The model's Bayesian Information Criterion (BIC)
+    ///
+    /// returns an f64 containing the BIC.
+    /// Is calculated during model training, useful for
+    /// determining optimal k for a given dataset.
+    pub fn bic(&self) -> f64 {
+        self.bic
     }
 
     fn membership_weights(&self, inputs: &Matrix<f64>) -> LearningResult<(Matrix<f64>, f64)> {
@@ -299,7 +311,7 @@ impl GaussianMixtureModel {
     fn update_params(&mut self, inputs: &Matrix<f64>, membership_weights: Matrix<f64>) {
         let n = membership_weights.rows();
         let d = inputs.cols();
-
+        let samples = inputs.rows() as f64;
         let sum_weights = membership_weights.sum_rows();
 
         self.mix_weights = &sum_weights / (n as f64);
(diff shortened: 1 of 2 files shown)


src/learning/gmm.rs
```diff
@@ -326,11 +338,30 @@ impl GaussianMixtureModel {
...
+    /// BIC = -2*log(l) + k * ln(n)
+    /// useful for determining the optimal number of clusters when iteratively generating GMMs.
+    /// log_lik = log likelihood criterion for the model, the calcaulated log_lik parameter is a sum so it needs to be divided by the total number of samples.
+    /// num_clusters = the number of clusters created in the model.
+    /// n = the total number of samples used to create the model.
+    fn calculate_bic(&self, n: f64) -> f64 {
+        let num_clusters:f64 = self.comp_count as f64;
+        let log_lik:f64 = self.log_lik / n;
+        let log_samples:f64 = n.ln();
+        assert!(!num_clusters.is_nan());
```
AtheMathmo: I think it is better that these are debug assertions (`debug_assert!`) - especially as currently they go in the inner loop.

src/learning/gmm.rs
```diff
@@ -326,11 +338,30 @@ impl GaussianMixtureModel {
...
+        assert!(!log_lik.is_nan());
+        assert!(!log_samples.is_nan());
+//        println!("num clusters: {} \t log_lik: {} \t log_samples: {}", num_clusters, log_lik, log_samples);
```
AtheMathmo: We should remove this comment (just for code hygiene)

src/learning/gmm.rs
```diff
@@ -326,11 +338,30 @@ impl GaussianMixtureModel {
             new_covs.push(cov_mat / sum_weights[k]);
 
         }
-
+        self.bic = self.calculate_bic(samples);
```
AtheMathmo: Do we need to do this within each `update_params` call? The `bic` function can only be called after training anyway so maybe we should simply store the number of samples as an `Option` field in the model?

This feels a little messy but may be worth it for the (perhaps minor) performance gain.

src/learning/gmm.rs
```diff
@@ -78,6 +80,8 @@ impl UnSupModel<Matrix<f64>, Matrix<f64>> for GaussianMixtureModel {
         // Initialization:
         let k = self.comp_count;
 
+        let variance = try!(inputs.variance(Axes::Row));
```
AtheMathmo: I don't think this line is doing anything?

src/learning/gmm.rs
```diff
@@ -105,9 +109,10 @@ impl UnSupModel<Matrix<f64>, Matrix<f64>> for GaussianMixtureModel {
         };
 
         self.model_covars = Some(vec![cov_mat; k]);
-
+//        let seed: &[usize] = &[1,2,3,4,5];
```
AtheMathmo: As before, should remove these before merging.

src/learning/gmm.rs
```diff
@@ -157,11 +161,32 @@ impl GaussianMixtureModel {
...
+            bic: 1f64,
             max_iters: 100,
             cov_option: CovOption::Full,
         }
     }
 
+    /// Constructs a Gaussian Mixture Model from components.
+    /// Useful for constructing a gmm from existing parts (IE json deserialization)
+    ///
+    pub fn deserialize(comp_count: usize,
```
AtheMathmo: I'm wary of introducing these manually like this. Instead we should try to use rustc-serialize or serde. Should be tackled separately in #122

src/learning/gmm.rs
```diff
@@ -257,17 +292,16 @@ impl GaussianMixtureModel {
 
         if let Some(ref covars) = self.model_covars {
             for cov in covars {
-                // TODO: combine these. We compute det to get the inverse.
+                // TODO: Fix this, this breaks stuff.
```
AtheMathmo: I'm not sure there is an alternative to computing the determinant and inverse here? At the least even a more numerically stable solution would retain some of the issues we see here.

src/learning/toolkit/rand_utils.rs
```diff
@@ -40,6 +40,41 @@ pub fn reservoir_sample<T: Copy>(pool: &[T], reservoir_size: usize) -> Vec<T> {
...
+/// ```
+/// use rand::{Rng, SeedableRng, StdRng};
+/// use rusty_machine::learning::toolkit::rand_utils;
+/// let seed: [_] = [1,2,3,4,5];
+/// let mut pool = &mut [1,2,3,4];
+/// let sample = rand_utils::reservoir_sample(pool, 3, seed);
+///
+/// println!("{:?}", sample);
+/// ```
+pub fn reservoir_sample_custom<T: Copy>(pool: &[T], reservoir_size: usize, seed: &[usize]) -> Vec<T> {
```
AtheMathmo: This is a welcome change in essence - though maybe should come in a separate PR - see #138 .

However this function signature should be modified to allow a generic type implementing the [`SeedableRng`](https://doc.rust-lang.org/rand/rand/trait.SeedableRng.html). Or perhaps just take as an argument an object implementing `Rng`?

AtheMathmo:
I'm closing this due to inactivity and the fact that it needs some work to be merge ready.

(Note that I'm happy to see this reopened with the relevant discussion/work in place. I just need to free up some mental capacity and tidy these PRs!)

Issue #150, opened 2016-10-08
Initialize GMM parameters with k-means


andrewcsmith:
The general practice seems to be to use GMM as an optimization of k-means. The initializer of GMM should therefore use k-means for the initial parameters, then GMM for fine-tuning.

AtheMathmo:
Yes this is indeed standard. I had thought about implementing this before but I'm not sure how exactly it should look. Certainly I'd like GMM to be accessible without the K-Means initialization. I see two approaches:

- Add some new initialization trait/enum similar to K-Means to GMM
- Allow a GMM model to be created `from_init_clusters`.

I think that it might even be worth implementing both of these - with the second suggestion coming first. The second would let us use K-Means to initialize GMM.

And to give a little more information on the K-Means initialization. The idea (from my knowledge) is to use K-Means to determine the location of the clusters (their means) and the data that belongs to them. We then compute the covariance matrix within each cluster (similarly to what we do now for the whole data set). This lets us assume a sensible mixture of gaussians over the whole data set which can be fine tuned.

---

## References

[Simple Methods for Initializing the EM Algorithm of GMMs](https://arxiv.org/pdf/1312.5946.pdf)
[Small (but englightening) SO question thread](http://stats.stackexchange.com/questions/163662/what-should-be-the-covariance-matrices-and-weights-for-initializing-em-gmm-with)

andrewcsmith:
My opinion for the API here is basically the same as it is in #153. We should create a `ClusterInitializer` trait that has whatever methods we need, and then `RandomInitializer` and `KMeansInitializer` structs to do the proper calculations accordingly.

andrewcsmith:
Same here—this issue is also fixed by #155.

PR #155, opened 2016-10-17
[WIP] Issue #152 use cholesky factorization

andrewcsmith:
This is basically a gigantic rewrite of the `GaussianMixtureModel` to make it use cholesky factorization instead of the inverse/determinant model. There's lots wrong with this commit stylistically, but for now I'm just hacking at it.

`cargo test -- --nocapture gmm_train`

I can't seem to figure out why the means of the model aren't separating. They all seem to drift in the same direction. Any thoughts? Anyway, if you know relatively quickly why that might be happening, I'd like to hear it.

### DIFF
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -37,6 +37,8 @@ use learning::{LearningResult, UnSupModel};
 use learning::toolkit::rand_utils;
 use learning::error::{Error, ErrorKind};
 
+const CONVERGENCE_EPS: f64 = 1.0e-15;
+
 /// Covariance options for GMMs.
 ///
 /// - Full : The full covariance structure.
@@ -92,7 +94,7 @@ impl UnSupModel<Matrix<f64>, Matrix<f64>> for GaussianMixtureModel {
 
             let (weights, log_lik_1) = try!(self.membership_weights(inputs));
 
-            if (log_lik_1 - log_lik_0).abs() < 1e-15 {
+            if (log_lik_1 - log_lik_0).abs() < CONVERGENCE_EPS {
                 break;
             }
 
--- a/src/learning/gmm.rs
+++ b/src/learning/gmm.rs
@@ -30,12 +30,18 @@
 //! // Probabilities that each point comes from each Gaussian.
 //! println!("{:?}", post_probs.data());
 //! ```
+extern crate rand;
+
 use linalg::{Matrix, MatrixSlice, Vector, BaseMatrix, BaseMatrixMut, Axes};
 use rulinalg::utils;
 
 use learning::{LearningResult, UnSupModel};
 use learning::toolkit::rand_utils;
 use learning::error::{Error, ErrorKind};
+
+use std::f64::consts::PI;
+use std::f64::EPSILON;
+use std::mem;
 
 const CONVERGENCE_EPS: f64 = 1.0e-15;
 
@@ -59,10 +65,15 @@ pub enum CovOption {
 #[derive(Debug)]
 pub struct GaussianMixtureModel {
     comp_count: usize,
+    // [n_features]
     mix_weights: Vector<f64>,
+    // [n_components, n_features]
     model_means: Option<Matrix<f64>>,
+    // n_components elements: [n_features, n_features]
     model_covars: Option<Vec<Matrix<f64>>>,
-    log_lik: f64,
+    // n_components elements: [n_features, n_features]
+    precisions_cholesky: Option<Vec<Matrix<f64>>>,
+    log_lik: Vector<f64>,
     max_iters: usize,
     /// The covariance options for the GMM.
     pub cov_option: CovOption,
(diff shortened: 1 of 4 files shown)


andrewcsmith:
I think it's getting closer. It doesn't work super well, but my guess is this is due to the initialization parameters and that's because it keeps getting stuck in a local minimum.

If you run the test a bunch of times, you see that occasionally the `log_lik` drops `< -2.0`, and the means start to look much more realistic.

I'll work on KMeans initialization, and hopefully that will help things.

andrewcsmith:
It's pretty clear that adding the k-means initialization step does not help things. 

```
initialized means:
⎡ 0.4962 -0.0198⎤
⎢-0.0147  0.5004⎥
⎣-0.5146 -0.5010⎦

# Final values
means:
⎡-0.0110 -0.0068⎤
⎢-0.0110 -0.0068⎥
⎣-0.0110 -0.0068⎦
log_lik:
-1.4304
cov:
⎡0.2872 0.0798⎤
⎣0.0798 0.2844⎦
cov:
⎡0.2872 0.0798⎤
⎣0.0798 0.2844⎦
cov:
⎡0.2872 0.0798⎤
⎣0.0798 0.2844⎦
test learning::gmm::gmm_train ... ok
```

For my particular application, I switched back to k-means at the moment. I'll look into this weird convergence bug again later, but for the next week or so I need to do other things.

AtheMathmo:
Thanks for your time on this. I'll take a look through the code and see if I can make sense of anything.

src/learning/gmm.rs
```diff
@@ -249,134 +292,290 @@ impl GaussianMixtureModel {
...
+    // == Returns
+    // weighted_log_prob : [n_features, n_components]
+    //
+    fn estimate_weighted_log_prob(&self, inputs: &Matrix<f64>) -> Matrix<f64> {
+        let mut log_prob = self.estimate_log_prob(&inputs);
+        // println!("log_prob: \n{:.4}", log_prob.select_rows(&[0, 1, 2, 3, 4, 5, 6]));
+        // println!("mix_weights: \n{:?}", &self.mix_weights);
+        let log_weights = self.mix_weights.iter().map(|w| w.ln());
+        for (lp, lw) in log_prob.iter_mut().zip(log_weights) {
+            *lp *= lw;
```
andrewcsmith: YES! Thanks.

andrewcsmith:
Just pushed the latest joy. It's good! It works quite well. The test uses k-means initialization by default, and I fixed a bug in the covariance calculation that sealed the deal.

It's still not ready to merge, because the design is pretty bad, but at least now there's an implementation to kick things off. I'll work on getting the gmm integration test to actually test things as well.

andrewcsmith:
I went ahead and improved a load of other things we discussed. It's pretty solid now. Still occasionally fails, but steadily increasing the regularization constant on each failure usually fixed that.

AtheMathmo:
I've had a brief look and things look very solid! I'll hopefully have time to properly review this in the next few hours.

Thanks for all the effort you've put into this, I'm really excited to check it out!

src/learning/gmm.rs
```diff
@@ -209,174 +212,404 @@ impl GaussianMixtureModel {
...
+    // == Paramers
+    // inputs : [n_samples, n_features]
+    //
+    //
+    fn estimate_log_prob(&self, inputs: &Matrix<f64>) -> Matrix<f64> {
+        let precisions: &Vec<Matrix<f64>> = self.precisions.as_ref().unwrap();
+        let ref model_means = self.model_means.as_ref().unwrap();
+        // The log of the determinant for each precision matrix
+        let log_det = precisions.iter().map(|m| m.det().ln());
+        // println!("log_det: {:?}", log_det);
```
AtheMathmo: Can we remove this commented line please?

_Edit: And the other commented `println!`s below_

src/learning/gmm.rs
```diff
@@ -209,174 +212,404 @@ impl GaussianMixtureModel {
...
+    fn update_gaussian_parameters(&mut self, inputs: &Matrix<f64>, resp: Matrix<f64>) {
+        self.mix_weights = resp.iter_rows()
+            .fold(Vector::new(vec![0.0; resp.cols()]), |mut acc, row| {
+                for (a, r) in acc.iter_mut().zip(row.iter()) {
+                    *a += *r;
                 }
-                Ok(cov_mat)
+                acc
+            })
+            + 10.0 * EPSILON;
```
AtheMathmo: This seems like black magic to me - I see it in the sklearn implementation but we should try to find some sources for it ourselves. Is it purely for numeric stability?

And a (very) minor point, I think it is better to `use f64` and here call `f64::EPSILON`. I think this makes the constant clearer, otherwise someone may think we have defined `EPSILON`.

src/learning/gmm.rs
```diff
@@ -30,75 +29,82 @@
...
-        self.model_covars = {
-            let cov_mat = try!(self.initialize_covariances(inputs, reg_value));
-            Some(vec![cov_mat; k])
-        };
+        // Initialize empty matrices for covariances and means
+        self.model_covars = Some(
+            vec![Matrix::zeros(inputs.cols(), inputs.cols()); k]);
+        self.model_means = Some(Matrix::zeros(inputs.cols(), k));
+
+        // Initialize responsibitilies and calculate parameters
```
AtheMathmo: Just noticed this typo, _responsibitilies_ -> 'responsibilities'

src/learning/gmm.rs
```diff
@@ -30,75 +29,82 @@
 //! // Probabilities that each point comes from each Gaussian.
 //! println!("{:?}", post_probs.data());
 //! ```
-use linalg::{Matrix, MatrixSlice, Vector, BaseMatrix, BaseMatrixMut, Axes};
-use rulinalg::utils;
+extern crate rand;
```
AtheMathmo: Shouldn't need this as we import the crate in `lib.rs`. Just `use rand` should work.

src/learning/gmm.rs
```diff
@@ -209,174 +212,404 @@ impl GaussianMixtureModel {
...
+            for col in 0..weighted_log_prob.cols() {
+                weighted_log_prob[[row, col]] -= log_prob_norm[row];
+            }
+        }
+
+        (log_prob_norm, weighted_log_prob)
+    }
+
+    fn update_gaussian_parameters(&mut self, inputs: &Matrix<f64>, resp: Matrix<f64>) {
+        self.mix_weights = resp.iter_rows()
```
AtheMathmo: Is [`sum_rows`](https://athemathmo.github.io/rulinalg/doc/rulinalg/matrix/slice/trait.BaseMatrix.html#method.sum_rows) not appropriate here?

src/learning/gmm.rs
```diff
@@ -209,174 +212,404 @@ impl GaussianMixtureModel {
...
+                *v /= *nk;
             }
         }
+
+        // Delegate updating the covariance matrices to the cov_option
+        let mut model_covars = self.model_covars.as_mut().unwrap();
+        self.cov_option.update_covars(model_covars, resp, inputs,
+                                      model_means, &self.mix_weights);
+
+        self.mix_weights /= inputs.rows() as f64;
```
AtheMathmo: This seems like it should be outside of this function.

AtheMathmo:
Actually I might have spoken too soon. I just ran a gmm simulation - sampling data from 3 gaussian clusters with specified mean and variances to see if it could recover the parameters. Sadly it consistently failed to do so, but not dramatically.

I'll try to get a PR in shortly to your branch to add this as an example in the repo - it's not the prettiest but seems it will be valuable for us developers and hopefully _some_ users.

andrewcsmith:
Cool. I didn't really do much in the way of a/b testing here. I'll see if I can tweak the parameters to get this to work better, or otherwise try to figure out what the problem is (probably next week though).

AtheMathmo:
Hopefully I'll have some time to dig into it in the next few days. It seems pretty minor so hopefully something will jump out.

src/learning/gmm.rs
```diff
@@ -209,174 +212,404 @@ impl GaussianMixtureModel {
...
+            let mut y = inputs * prec;
+            // println!("y: \n{:.4}", &y.select_rows(&[0, 1, 2, 3, 4, 5, 6]));
+            // Matrix of shape [1, n_features]
+            let z: Matrix<f64> = model_means.select_rows(&[k]) * prec;
+            // println!("z: \n{:.4}", &z);
+            
+            // Subtract the mean of each column from the matrix y
+            for col in 0..y.cols() {
+                for row in 0..y.rows() {
+                    y[[row, col]] -= z[[0, col]];
```
AtheMathmo: Minor stuff here - this access pattern will incur some cache misses.

We should swap the two loops, so that we access contiguous data in sequence (right now we jump down the columns first). We should also probably use `get_unchecked_mut` here, so remove the bound checks that we know we don't need.

AtheMathmo:
I finally got a little time to look at this. I think I checked over everything and compared to the scikit implementation - sadly I couldn't spot anything incorrect. It is failing even in the 1d case (as in the example added) so I don't think it is all of the Cholesky stuff. I'll try to find some time to dig around a little more, but we may just have to step through and compare to a working implementation to spot where things go wrong...

andrewcsmith:
I bounced back and forth between the `CholeskyFull` and `Diagonal` interpretations, and while they're both wrong it seems like the `CholeskyFull` has generally higher values for the variance, although it also tends to overestimate the variance of the point at a mean of `0.0`.

I'm getting fairly consistent results, but they're just not particularly correct. I believe it's external to the `impl CovOption` blocks. My guess is that it's somewhere in the update code. It seems that the variances are wildly off, which leads me to believe that the calculation of the model responsibilities is off. That's just a guess. 

(I'm only commenting because I've been poking around for an hour or so and want to save these notes before moving on with my life. Sorry if it's not particularly helpful.)

AtheMathmo:
Sat down with this again today.

I've been unable to track down the issue, but agree with you that it is probably the update (or atleast gaussian estimation) code.

I think the best approach to find the issue is to write unit tests for each part of the process using the scikit implementation to help. Basically check that each function is computing what we expect with some dummy input.

The good new is that the new `assert_matrix_eq!` macro should make this a lot easier to do. I'd like to sit down and get this done but I'm not sure when I'll have time. Maybe by next weekend I could take a look...

AtheMathmo:
I'm going to try to tackle this again when https://github.com/AtheMathmo/rulinalg/pull/150 is merged.

This should make this work a little easier to complete.

PR #157, opened 2016-12-06
Add confusion_matrix

theotherphil:
Definitely needs clearer documentation, and a module that currently has only one function (with the same name as the module) might a bit confusing. Look reasonable otherwise? I also plan on adding normalised_confusion_matrix which reports fractions of total samples instead of counts.

### DIFF
--- a/src/analysis/confusion_matrix.rs
+++ b/src/analysis/confusion_matrix.rs
@@ -4,18 +4,21 @@ use std::hash::Hash;
 use std::collections::HashMap;
 use linalg::Matrix;
 
-/// Returns a matrix C where C_ij is the count of the samples which were
-/// predicted to lie in class j but actually lie in class i. (i.e.
-/// the row is the true class and the column is the predicted.).
+/// Returns a square matrix C where C_ij is the count of the samples which were
+/// predicted to lie in the class with jth label but actually lie in the class with
+/// ith label.
 ///
 /// # Arguments
 /// * `predictions` - A series of model predictions.
-/// * `targets` - A slice of equal length to predictions, containing the target results.
-/// * `labels` - If not provided then the returned matrix rows and columns correspond to
-///              the distinct elements of union(predictions, targets), in increasing
-///              order. If labels are provided then these are used instead, in the order
-///              provided. Note that this means that the confusion matrix will only contain
-///              entries for the elements of labels.
+/// * `targets`     - A slice of equal length to predictions, containing the
+///                   target results.
+/// * `labels`      - If None then the rows and columns of the returned matrix
+///                   correspond to the distinct labels appearing in either
+///                   predictions or targets, in increasing order.
+///                   If Some then the rows and columns correspond to the provided
+///                   labels, in the provided order. Note that in this case the
+///                   confusion matrix will only contain entries for the elements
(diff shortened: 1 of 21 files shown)


src/analysis/confusion_matrix.rs
```diff
@@ -0,0 +1,182 @@
...
+    use linalg::Matrix;
+
+    #[test]
+    fn confusion_matrix_no_labels() {
+        let truth       = vec![2, 0, 2, 2, 0, 1];
+        let predictions = vec![0, 0, 2, 2, 0, 2];
+
+        let confusion = confusion_matrix(&predictions, &truth, None);
+
+        let expected = Matrix::new(3, 3, vec![
```
AtheMathmo: In rulinalg we're trying to follow the convention of using the new [`matrix!`](https://athemathmo.github.io/rulinalg/doc/rulinalg/macro.matrix.html) macro for any new tests. It's a minor point but I thought I'd bring it up as it is a _little_ bit more readable than the `Matrix::new` function with a `vec!` macro.

src/analysis/confusion_matrix.rs
```diff
@@ -0,0 +1,182 @@
...
+/// ```
+/// use rusty_machine::analysis::confusion_matrix::confusion_matrix;
+/// use rusty_machine::linalg::Matrix;
+///
+/// let truth       = vec![2, 0, 2, 2, 0, 1];
+/// let predictions = vec![0, 0, 2, 2, 0, 2];
+///
+/// let confusion = confusion_matrix(&predictions, &truth, None);
+///
+/// let expected = Matrix::new(3, 3, vec![
```
AtheMathmo: We could also use the `matrix!` macro here - though I think it would be better to modify doc-facing code all at once.

theotherphil:
Thanks for the comments. They all sound reasonable. I'm off for a few days now without a computer but I'll make the suggested changes next week.

The way labels work is copied from scikit learn. I'll make the docs a bit clearer.

AtheMathmo:
>  I'll make the suggested changes next week.

Great! Take your time :)

> The way labels work is copied from scikit learn. I'll make the docs a bit clearer.

Ah ok - as you can tell I've never used this from scikit learn before :). I think it does make sense and as others may be familiar with the approach already I don't think you need to worry too much about the docs.

AtheMathmo:
This looks good to me now!

Will have time to give it one last proper look over tomorrow morning. Will hopefully get this in then.

theotherphil:
I've tweaked the docs and switched the tests to using the matrix! macro. I'm not sure about taking an Into<Vec> - it feels a bit of a waste to allocate new vecs if the user already has a contiguous chunk of memory to hand, which I'm assuming will be the typical case.

I'll change that if you have a strong opinion on the matter, but otherwise I think this should be mergeable now.

(Got distracted writing this comment and now I see you've already commented on the latest changes!)

AtheMathmo:
> it feels a bit of a waste to allocate new vecs if the user already has a contiguous chunk of memory to hand

I see what you mean. I don't feel strongly about it and will be happy to merge without this.

AtheMathmo:
Sorry it took me a while to get around to this, merging now!

PR #161, opened 2016-12-16
ENH: Add datasets

### LINKED ISSUE #115: Wrappers for standard datasets

sinhrks:
Closes #115. Added ``Dataset`` struct which has ``data()`` and ``target()`` impl (intended for supervised learning).

Adding more data once API looks OK.


src/datasets/mod.rs
```diff
@@ -0,0 +1,195 @@
...
+
+
+/// Load iris dataset.
+///
+/// The data set contains 3 classes of 50 instances each, where each class refers to a type of iris plant.
+///
+/// Lichman, M. (2013). UCI Machine Learning Repository [http://archive.ics.uci.edu/ml].
+/// Irvine, CA: University of California, School of Information and Computer Science.
+pub fn load_iris() -> Dataset<Matrix<f64>, Vector<usize>> {
+    let data: Vec<f64> = vec![5.1, 3.5, 1.4, 0.2,
```
AtheMathmo: Minor: It might be easier to use the `matrix!` macro here. My thinking is that if we need to add a row there's a little less work.

src/datasets/mod.rs
```diff
@@ -0,0 +1,195 @@
+use std::fmt::Debug;
+
+use rulinalg::matrix::Matrix;
+use rulinalg::vector::Vector;
+
+/// Dataset container
+#[derive(Clone, Debug)]
+pub struct Dataset<D, T> where D: Clone + Debug, T: Clone + Debug {
```
AtheMathmo: I think this makes sense for now. We might want to be more strict in future if we want to be generic over `DataSets`. However, this is something that I don't think we will ever want to do.

src/lib.rs
```diff
@@ -219,3 +219,6 @@ pub mod analysis {
     pub mod cross_validation;
     pub mod score;
 }
+
+/// Module for datasets.
+pub mod datasets;
```
AtheMathmo: We should feature gate this. My thinking is that if we have a few datasets users will not want to download all of this data by default.

To do this:

- Add a new feature to [`Cargo.toml`](https://github.com/AtheMathmo/rusty-machine/blob/c7dcaaf0daf867b8f0d217f81a81c67d654b8b87/Cargo.toml#L14)
- In [`lib.rs` add a feature flag](https://github.com/AtheMathmo/rusty-machine/blob/3ab5a0513d4fc4aa7a47cd7c97c280cff66902f2/src/lib.rs#L210-L212)

sinhrks:
Added feature gates. Is it ok to be included by default ATM (as it is likely to be used in most tests)?

AtheMathmo:
Thanks for the update.

It looks good but I'm a little cautious about having the datasets flag included by default. I wanted it feature flagged specifically so that it had to be _opted-in_. I can see that we will probably want to use it in some tests but I'd try a few ways around this first.

- Can we fit all tests in the `datasets` module?
- Can we somehow put the tests in the `tests` folder behind the feature flag too?
- As a last resort, do as you have done and keep the "datasets" flag for the "test" feature.

Finally note that we will need to modify the travis CI matrix to include the "datasets" flag.

sinhrks:
OK, made ``datasets`` as optional, and fixed travis tests.

AtheMathmo:
This looks good to me but before merging I'd like to check out the branch and play around with it a little.

Thanks!

AtheMathmo:
I checked out the code and I have a few thoughts. I am happy to merge this in without any further changes but we should at least write up a tracking issue for improvements.

I think the description of `load_iris` should include more information on the dataset. Or a link directly to the iris dataset: http://archive.ics.uci.edu/ml/datasets/Iris (instead of just the `/ml` root). We should say what the attributes and classes are. From the link:

> Attribute Information:
> 
> 1. sepal length in cm 
> 2. sepal width in cm 
> 3. petal length in cm 
> 4. petal width in cm 
> 5. class: 
> -- Iris Setosa 
> -- Iris Versicolour 
> -- Iris Virginica

Also I think it might be a good idea to organize the datasets module a little differently. If we add more datasets the module is going to get large quickly and difficult to manage. I think we should move the iris data into a new `iris` module. This `iris` module should have a `pub fn load() -> DataSet`. As a user I would be quite happy to do:

```rust
use rusty_machine::datasets:iris;

let (inputs, targets) = iris::load();
```

But we could also use the current format and have a `load_iris` function in the `datasets` module that calls `iris::load`. In this case we might want to keep the `iris` module private?

Let me know what you think. If you don't want to make any of these changes now I'll merge and move this information to a separate ticket.

sinhrks:
Thx for the comment. I've did a change requested. Pls take a look when u have a time.

Cargo.toml
```diff
@@ -11,7 +11,10 @@ readme = "README.md"
 license = "MIT"
 
 [features]
+default = []
 stats = []
+datasets = []
+test = []
```
AtheMathmo: We don't need to include the `test` or `default` features. These already exist as defined here.

src/datasets/iris.rs
```diff
@@ -0,0 +1,191 @@
+use rulinalg::matrix::Matrix;
+use rulinalg::vector::Vector;
+
+use super::Dataset;
+
+/// Load iris dataset.
```
AtheMathmo: This description is great!

Cargo.toml
sinhrks: Correct, removed.

AtheMathmo:
Thank you! Merging now.

PR #162, opened 2016-12-20
ENH: implement Agglomerative (hierarchical) clustering

sinhrks:
This needs a discussion, because hierarchical clustering uses the same data for training and prediction. It doesn't meet very well with current ``SupModel`` trait.

Maybe we should have ``train_predict`` method to use same data for training and prediction?


AtheMathmo:
Awesome! Thank you so much for these PRs - sorry that it's taking me a little while to look through them properly as well. This is another algorithm I'm not super familiar with and so will need a little time to understand it.

There are definitely some issues with the current traits for learning. We've had discussions ( #124 ) about these before but it is hard to agree on the best. If you have anything to add I would love to hear your opinion.

As for this particular example, could you elaborate a little more on what exactly is lacking in the trait signature. From a brief look it seems that you train the model and return the clusters. The way I have handled this elsewhere is to use the `UnSupModel` trait and store the clusters within the model, providing an access function. And then the predict function will assign the new points to their closest cluster. This is a little messy but keeps things consistent until we can find a nicer way to link models and traits together.

src/learning/agglomerative.rs
```diff
@@ -0,0 +1,415 @@
...
+        }
+    }
+}
+
+/// Distance Matrix
+#[derive(Debug)]
+struct DistanceMatrix {
+    // Distance is symmetric, no need to hold all pairs
+    // use HashMap to easier update
+    data: HashMap<(usize, usize), f64>
```
AtheMathmo: This  comment is more to open up a conversation - are you sure this is necessary?

We are storing 3 64bit values for half the entries of our original matrix. This means we take up more memory than just using an abstraction over the original matrix. Of course - if we need to regularly use the distance then it is more efficient to store the computation once. In this case maybe we can just store the values in a `Vec` and implement some custom indexing?

This is all pretty minor (but interesting) stuff. I imagine it will probably end up being kept as-is.

src/learning/agglomerative.rs
```diff
@@ -0,0 +1,415 @@
...
+    /// Create distance matrix fron input matrix
+    fn from_mat(inputs: &Matrix<f64>) -> Self {
+        assert!(inputs.rows() > 0, "input is empty");
+
+        let n = inputs.rows() - 1;
+        let mut data: HashMap<(usize, usize), f64> = HashMap::with_capacity(n * n);
+
+        unsafe {
+            for i in 0..n {
+                for j in i..inputs.rows() {
```
AtheMathmo: We can skip the `j == i` case here as we tackle it in the `get` function below.

src/learning/agglomerative.rs
```diff
@@ -0,0 +1,415 @@
...
+            *self.data.get(&(j, i)).unwrap()
+        } else {
+            *self.data.get(&(i, j)).unwrap()
+        }
+    }
+
+    /// Add distance between i-th and j-th item
+    /// i must be smaller than j
+    fn insert(&mut self, i: usize, j: usize, dist: f64) {
+        assert!(i < j, "i must be smaller than j");
```
AtheMathmo: I would make this `assert_debug`. The function is only used internally so we wouldn't expect a failure - as such we don't want a performance hit when run in release mode.

src/learning/agglomerative.rs
```diff
@@ -0,0 +1,415 @@
...
+        self.data.insert((i, j), dist);
+    }
+
+    /// Delete distance between i-th and j-th item
+    fn delete(&mut self, i: usize, j: usize) {
+        assert!(i != j, "DistanceMatrix doesn't store distance when i == j, because it is 0.0");
```
AtheMathmo: Same here, switch to `assert_debug`.

src/learning/agglomerative.rs
```diff
@@ -0,0 +1,415 @@
...
+
+use std::collections::{BTreeMap, HashMap};
+use std::f64;
+
+use linalg::{Matrix, BaseMatrix, Vector};
+use learning::{LearningResult};
+
+/// Agglomerative clustering distances
+#[derive(Debug)]
+pub enum Metrics {
```
AtheMathmo: I'm not sure if this is relevant here but I would consider using a trait here instead of an enum. The upside is that users can define their own metrics and plug them into the model (without modifying rusty-machine source code). The downside is an extra generic parameter on the model.

sinhrks:
You're right. Renamed ``Metrics`` to ``Linkage``.

src/learning/agglomerative.rs
sinhrks: This is for simple implementation rather than memory efficiency. 

Because distance matrix should keep only remaining nodes and merged clusters, mapping node id to matrix cell (of triangular flattened to vector) makes impl complex. 

I think it is OK ATM as hierarchical clustering is not likely to be used for larger data, but appreciated if anyone lmk better impl.

AtheMathmo:
Sorry I forgot to comment on this again.

Thanks for making those changes - I think that this is probably ready to be merged now. Before I do merge I'd like to have a play with it and convince myself it is working as expected. Do you have any recommendations on how I can do this? Any data sets that lend themselves to this kind of model?

sinhrks:
iris is ok as test data. Here is R example( https://cran.r-project.org/web/packages/dendextend/vignettes/Cluster_Analysis.html). I'm adding tests once #161 has been merged.

Note: current test data is derived from a Japanese textbook in R.

Another point is training API. Currently, it doesn't impl ``UnSupModel``. What do you think to change ``UnSupModel`` to:

```
    pub trait UnSupModel<T, U> {
        /// Predict output from inputs.
        fn predict(&self, inputs: &T) -> LearningResult<U>;

        /// Train the model using inputs.
        fn train(&mut self, inputs: &T) -> LearningResult<()>;

        /// Train the model using inputs, then predict outputs from the same inputs.
        fn train_predict(&mut self, inputs: &T) -> LearningResult<U> {
            self.train(inputs).unwrap();
            self.predict(inputs)
    }
```

``AgglomerativeClustering`` should raise ``unimplemented!`` in ``predict/train``. And should have custom impl for ``train_predict``.

PR #165, opened 2016-12-26
LinRegressor to use solve rather than inverse

sinhrks:
related to #152. Change ``LinRegressor`` to use ``.solve`` rather than ``.inverse``.


AtheMathmo:
I am happy to merge this in as it is. However, I wasn't sure whether you would want to move the trees dataset into the new `datasets` module?

Let me know what you'd prefer.

sinhrks:
OK, added ``trees`` dataset. Also, fixed conditional compile flag for datasets tests because it is not executed properly.

src/learning/lin_reg.rs
```diff
@@ -87,11 +87,8 @@ impl SupModel<Matrix<f64>, Vector<f64>> for LinRegressor {
         let full_inputs = ones.hcat(inputs);
 
         let xt = full_inputs.transpose();
-
-        self.parameters =
-            Some(((&xt * full_inputs).inverse().expect("Could not compute (X_T X) inverse.") *
-                  &xt) * targets);
-
+        self.parameters = Some((&xt * full_inputs).solve(&xt * targets)
+                                                  .expect("Could not solve equation."));
```
AtheMathmo: I think that this error message could be a little clearer. Maybe: `Unable to solve linear equation`? Even this could be improved I'm sure.

sinhrks: ok, updated the message.

AtheMathmo:
Thank you for this!

I'm sure there are lots of other areas where numerical stability (and accuracy) can be improved. A big one is the GLM module (see #152). Just in case you want to try your hand at that!

PR #166, opened 2016-12-27
Add Transformer.fit

sinhrks:
Based on the discussion in #158, added ``.fit`` method to ``Transformer``. This also allows basic transformers to fit to training data and transform test data.

Needs to decide:

- ``Shuffler`` doesn't needs to be ``fit``. There is an option to split ``Transformer`` depending on needs to fit, but I feel it is complex rather than convenient from user's point of view.
- To keep backward compat, ``.transform`` internally calls ``.fit`` if it is not fitted yet. Or should this be breaking change to force users to call ``.fit`` first.
- Is it ok to change ``MinMaxScaler`` to store trained data as ``Vector`` rather than ``Vec``, which should be compat with others.

Once decisions are made, let me add tests to validate changes.

### DIFF
--- a/src/data/transforms/minmax.rs
+++ b/src/data/transforms/minmax.rs
@@ -135,18 +135,9 @@ impl<T: Float> Transformer<Matrix<T>> for MinMaxScaler<T> {
     }
 
     fn transform(&mut self, mut inputs: Matrix<T>) -> Result<Matrix<T>, Error> {
-        match (&self.scale_factors, &self.const_factors) {
+        if let (&None, &None) = (&self.scale_factors, &self.const_factors) {
             // if Transformer is not fitted to the data, fit for backward-compat.
-            (&None, &None) => {
-                let res = self.fit(&inputs);
-                match res {
-                    Err(err) => {
-                        return Err(err);
-                    },
-                    _ => {}
-                }
-            },
-            _ => {}
+            try!(self.fit(&inputs));
         }
 
         if let (&Some(ref scales), &Some(ref consts)) = (&self.scale_factors, &self.const_factors) {
--- a/src/data/transforms/minmax.rs
+++ b/src/data/transforms/minmax.rs
@@ -25,7 +25,7 @@
 //! ```
 
 use learning::error::{Error, ErrorKind};
-use linalg::{Matrix, BaseMatrix, BaseMatrixMut};
+use linalg::{Matrix, BaseMatrix, BaseMatrixMut, Vector};
 use super::{Invertible, Transformer};
 
 use rulinalg::utils;
@@ -42,9 +42,9 @@ use libnum::Float;
 #[derive(Debug)]
 pub struct MinMaxScaler<T: Float> {
     /// Values to scale each column by
-    scale_factors: Option<Vec<T>>,
+    scale_factors: Option<Vector<T>>,
     /// Values to add to each column after scaling
-    const_factors: Option<Vec<T>>,
+    const_factors: Option<Vector<T>>,
     /// The min of the new data (default 0)
     scaled_min: T,
     /// The max of the new data (default 1)
@@ -82,9 +82,12 @@ impl<T: Float> MinMaxScaler<T> {
 }
 
 impl<T: Float> Transformer<Matrix<T>> for MinMaxScaler<T> {
-    fn transform(&mut self, mut inputs: Matrix<T>) -> Result<Matrix<T>, Error> {
+
+    fn fit(&mut self, inputs: &Matrix<T>) -> Result<(), Error> {
         let features = inputs.cols();
 
+        // ToDo: can use min, max
+        // https://github.com/AtheMathmo/rulinalg/pull/115
         let mut input_min_max = vec![(T::max_value(), T::min_value()); features];
 
         for row in inputs.iter_rows() {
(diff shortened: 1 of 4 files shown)


src/data/transforms/minmax.rs
```diff
@@ -82,9 +82,12 @@ impl<T: Float> MinMaxScaler<T> {
 }
 
 impl<T: Float> Transformer<Matrix<T>> for MinMaxScaler<T> {
-    fn transform(&mut self, mut inputs: Matrix<T>) -> Result<Matrix<T>, Error> {
+
+    fn fit(&mut self, inputs: &Matrix<T>) -> Result<(), Error> {
         let features = inputs.cols();
 
+        // ToDo: can use min, max
+        // https://github.com/AtheMathmo/rulinalg/pull/115
```
AtheMathmo: Just to note that this implementation may actually be more efficient as we handle it on one pass of the rows.

Certainly worth checking though (besides - we will have some other breaking changes by the time this PR lands).

src/data/transforms/minmax.rs
```diff
@@ -130,28 +129,55 @@ impl<T: Float> Transformer<Matrix<T>> for MinMaxScaler<T> {
...
 
-            utils::in_place_vec_bin_op(row, &consts, |x, &y| {
-                *x = *x + y;
-            });
+    fn transform(&mut self, mut inputs: Matrix<T>) -> Result<Matrix<T>, Error> {
+        match (&self.scale_factors, &self.const_factors) {
+            // if Transformer is not fitted to the data, fit for backward-compat.
+            (&None, &None) => {
+                let res = self.fit(&inputs);
+                match res {
```
AtheMathmo: Can remove this match block by using `try!(self.fit(&inputs))`.

src/data/transforms/minmax.rs
```diff
@@ -130,28 +129,55 @@ impl<T: Float> Transformer<Matrix<T>> for MinMaxScaler<T> {
...
+        self.scale_factors = Some(Vector::new(scales));
+        self.const_factors = Some(Vector::new(consts));
+        Ok(())
+    }
 
-            utils::in_place_vec_bin_op(row, &consts, |x, &y| {
-                *x = *x + y;
-            });
+    fn transform(&mut self, mut inputs: Matrix<T>) -> Result<Matrix<T>, Error> {
+        match (&self.scale_factors, &self.const_factors) {
```
AtheMathmo: I think using [`if let ...`](https://doc.rust-lang.org/book/if-let.html) would be a little clearer here.

src/data/transforms/mod.rs
```diff
@@ -21,6 +21,8 @@ pub use self::standardize::Standardizer;
 
 /// Trait for data transformers
 pub trait Transformer<T> {
+    /// Fit Transformer to input data, and stores the transformation in the Transformer
+    fn fit(&mut self, inputs: &T) -> Result<(), error::Error>;
     /// Transforms the inputs and stores the transformation in the Transformer
```
AtheMathmo: I would add to this comment and state that if this function is used without `fit`ting first then the `Transformer` will call `fit` itself.

src/data/transforms/shuffle.rs
```diff
@@ -74,6 +75,12 @@ impl Default for Shuffler<ThreadRng> {
 ///
 /// Under the hood this uses a Fisher-Yates shuffle.
 impl<R: Rng, T> Transformer<Matrix<T>> for Shuffler<R> {
+
+    #[allow(unused_variables)]
+    fn fit(&mut self, inputs: &Matrix<T>) -> Result<(), Error> {
+        unimplemented!();
```
AtheMathmo: This is a tricky one.

I think we would prefer to just have the function be a no-op. My only concern would be that users might expect to be able to retrieve information about the transformation after `fit`ting - for example a list of row-indices to be swapped.

Otherwise without any additional documentation informing them of the panic I think almost all users would be confused and tripped up by this.

src/data/transforms/mod.rs
sinhrks: A point is whether calling ``transform`` before ``fit`` is allowed  as standard behavior.

I'm + 1 to remove it in future. thus better to show warning and do not describe the behavior in the doc?

AtheMathmo:
This was originally a comment reply but I've moved it here so it can persists.

If we want to disallow using `transform` without `fit` then I believe the best way is to control this with the traits. I.e. add a new trait called `TransformFitter` (or something better named). Which has the following signature (or similar):

```rust
pub trait TransformFitter<U, T: Transformer> {
    fn fit(self, inputs: U) -> T;
}
```

We can then make it so that any `Transformer`s which require a `fit` before use cannot be directly instantiated (no constructor functions, and private fields). If a transform does not need a fit (like `Shuffler`) then we can give it constructor functions instead of a `TransformFitter` struct.

Another advantage of this approach is that we do not have to deal with `Option` fields all the time.

sinhrks:
Splitting traits sounds good. Maybe it should be done at the same time with #124? 

Any change is required for the PR atm?

AtheMathmo:
As this PR introduces breaking changes anyway I think there isn't much reason to delay the change?

I'll read through the PR again properly soon - am a little tied up at the moment.

sinhrks:
@AtheMathmo I don't think the PR breaks something ATM, because all existing tests are passed. Users can call ``transform`` directly as the same as previous versions.

AtheMathmo:
Ah yes, I see your point! I'll take one last look through and hopefully get this merged.

PR #167, opened 2016-12-27
Bump to rulinalg 0.4

sinhrks:


AtheMathmo:
Thanks for doing this! If we merge we need to do a bump to 0.6 due to breaking changes (we reexport much of rulinalg). Because of this I'll try to get some of the other pending PRs in first.

AtheMathmo:
I'm in a tricky situation with this PR. We have a few other big PRs pending (as you can see, you authored many of them!). I would like to get this merged in first, but it puts pressure on other PRs to accommodate the breaking changes.

What do you think of this? If you think we should get this in first then I am happy to make the changes to this PR myself and get it merged.

AtheMathmo:
Closing in favour of #174

Issue #145, opened 2016-09-25
Improvements to linear models


AtheMathmo:
The current techniques used in the linear models are sub-par and numerically unstable. There are better techniques available that we should use. This is most notably an issue for the GLM module - but improvements can be made in the linear model and logistic model as well (which are of course special cases).

The current implementations are as follows:
- Linear regression : compute `(X.t * X)^-1 * X.t * y`
- Logistic regression : we use gradient descent in the `logistic_reg` module
- General Linear Models : we use a naive IRLS implementation which again computes inverses directly.

We should focus on improving GLM's as the issues crop up very often here. The [GLMs, abridged](http://bwlewis.github.io/GLM/) article provides a great overview of GLMs in R - we can make good use of this as a resource for rusty-machine. Rulinalg provides all the techniques we need - though of course without the performance given by BLAS/LAPACK.

For linear regression we can also make use of better techniques to solve the least squares problem. More research needed here.

Finally we should add Regularization to the linear models. The `learning::toolkit::regularization` model should do most of the work for us - though this was designed for neural nets and may need some changes.

angusturner:
I'd like to help! 

I'm quite new to ML, but would like to have a stab at Levenberg-Marquardt optimisation (or similar). At a glance, the maths doesn't seem *too difficult*, so if this sounds like a useful contribution I'll get started on it. From what I understand, this could significantly outperform vanilla GD on linear and logistic regression (and neural net).

AtheMathmo:
Thank you for your comment! I hadn't heard of the Levenberg-Marquardt algorithm before you suggested it here. It looks like it could be a valuable addition to the linear and logistic regression models. However from my own reading I'm not sure whether this would be a stable alternative in some of the more tricky GLMs (as described in the [aforementioned article](http://bwlewis.github.io/GLM/)). That said - if you're happy to try and get this implemented I would be glad to review it!

As a final thought. It might be worth trying to be generic over the underlying algorithm. This would make it easy for us to compare them across different models and provide options to our users for those cases where different algorithms are needed. This is probably quite a big task which we can put off for now.

phasedchirp:
[...]

PR #169, opened 2016-12-30
ENH: Add Ridge Regression

sinhrks:
Added ``RidgeRegressor`` under ``lin_reg``.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ It aims to combine speed and ease of use - without requiring a huge number of ex
 This project began as a way for me to learn Rust and brush up on some less familiar machine learning algorithms and techniques.
 Now the project aims to provide a complete, easy to use, machine learning library for Rust.
 
-This library is still very much in early stages of development. Although there are a good number of algorithms many other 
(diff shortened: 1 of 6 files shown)


AtheMathmo:
Thanks for opening this!

I originally had plans to include this using the [`regularization`](https://athemathmo.github.io/rusty-machine/doc/rusty_machine/learning/toolkit/regularization/index.html) module. I was going to add this as part of the `LinRegressor` model but I don't actually mind the idea of having a new model as you have here.

Perhaps we could have a core model which uses the `Regularization` enum - and then abstractions like, `LinRegressor`, [`LassoRegressor`](https://en.wikipedia.org/wiki/Lasso_(statistics)), `RidgeRegressor`?

I will take a proper look through this PR soon.

AtheMathmo:
You may also want to check #145 which discusses this a little and more.

AtheMathmo:
Have you had a chance to think about my comments above?

AtheMathmo:
I'm closing this as it seems to have stagnated from the comments above. If you would like to reopen it please don't hesitate! I'd like to talk this through a little more.

Issue #163, opened 2016-12-25
Add Basic Transformers


sinhrks:
Related to #158. I think below transformers are useful.

- min / max scaling (scaling inputs to [0, 1] using min/max)
- normalize (each rows are rescaled to have the same norm)
- standardize (transform to zero-mean and unit-variance)

AtheMathmo:
Quite a few of these [already exist](https://athemathmo.github.io/rusty-machine/doc/rusty_machine/data/transforms/index.html).

I'd welcome a PR to add a normalized transform :)

sinhrks:
ah, haven't noticed. gr8:)

AtheMathmo:
I am going to close this @sinhrks . Feel free to reopen an issue for the `normalize` transformer.

PR #170, opened 2017-01-02
ENH: Add Normalizer

sinhrks:
Add normalize transformer (#163). Because normalizer has no parameters to fit, it is separate from #166.

### DIFF
--- a/src/data/transforms/normalize.rs
+++ b/src/data/transforms/normalize.rs
@@ -3,7 +3,9 @@
 //! This module contains the `Normalizer` transformer.
 //!
 //! The `Normalizer` transformer is used to transform input data
-//! so that the l2 norm of each rows are all equal to 1.
+//! so that the norm of each row is equal to 1. By default the
+//! `Normalizer` uses the `Euclidean` norm.
+//!
 //! If input data has a row with all 0, `Normalizer` keeps the row as it is.
 //!
 //! Because transformation is performed per row independently,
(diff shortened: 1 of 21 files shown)


src/data/transforms/normalize.rs
```diff
@@ -0,0 +1,158 @@
+//! The Normalizing Transformer
+//!
+//! This module contains the `Normalizer` transformer.
+//!
+//! The `Normalizer` transformer is used to transform input data
+//! so that the l2 norm of each rows are all equal to 1.
+//! If input data has a row with all 0, `Normalizer`` keeps the row as it is.
```
AtheMathmo: Extra '`' here

src/data/transforms/normalize.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// let _ = Normalizer::new();
+    /// ```
+    pub fn new() -> Self {
+        Normalizer::default()
+    }
+}
+
+impl<T: Float> Transformer<Matrix<T>> for Normalizer {
+    fn transform(&mut self, mut inputs: Matrix<T>) -> Result<Matrix<T>, Error> {
+        let dists: Vec<T> = inputs.iter_rows().map(|x| dist(x)).collect();
```
AtheMathmo: I believe we can do this more efficiently than the `dist` function. Vectorizing the dot product takes some work, which is why we have the [`utils::dot`](https://athemathmo.github.io/rulinalg/doc/rulinalg/utils/fn.dot.html) function in rulinalg.

Once we merge rulinalg 0.4.x we can update to using the `MatrixNorm` trait directly here.

Oh, and just in case I am wrong and the `dist` function is better. This line can be written as:

```rust
let dists: Vec<T> = inputs.iter_rows().map(dist).collect();
```

sinhrks:
Updated to meet with #166.

src/data/transforms/normalize.rs
```diff
@@ -0,0 +1,164 @@
...
+///
+/// The Normalizer provides an implementation of `Transformer`
+/// which allows us to transform the all rows to have the same l2 norm (1).
+///
+/// See the module description for more information.
+#[derive(Debug)]
+pub struct Normalizer;
+
+/// Create a `Normalizer`.
+impl Default for Normalizer {
```
AtheMathmo: This can be derived on the above struct (though I don't think even this is necessary).

```rust
let a = Normalizer;
let b = Normalizer::default();
let c = Normalizer::new();
```

These all do the same thing here. I'd suggest we only keep the first.

AtheMathmo:
Closing for #175 .

PR #172, opened 2017-01-24
Linear Dirichlet Allocation

dyule:
I've created a basic implementation of linear dirichlet allocation.  Unlike the other learning algorithms, it's unsupervised, so the training method is just empty.  It doesn't support non-symmetric parameters, but that could fairly easily be added in.  I've tried to document everything clearly.

However I built this with v4 of rulinalg, so this pull request includes #167.

I'm happy to take any feedback on the approach here, and update as necessary.

### DIFF
--- a/examples/lda_gen.rs
+++ b/examples/lda_gen.rs
@@ -1,0 +1,125 @@
+extern crate rusty_machine;
+extern crate rand;
+extern crate rulinalg;
+
+use rusty_machine::linalg::{Matrix, BaseMatrix, Vector};
+use rusty_machine::learning::UnSupModel;
+use rusty_machine::learning::lda::LDA;
+
+use rand::{thread_rng, Rng};
+use rand::distributions::{gamma, IndependentSample};
+
+/// Given `topic_count` topics, this function will create a distrbution of words for each
+/// topic.  For simplicity, this function assumes that the total number of words in the corpus
+/// will be `(topic_count / 2)^2`.
+fn generate_word_distribution(topic_count: usize) -> Matrix<f64> {
+    let width = topic_count / 2;
+    let vocab_size = width * width;
+    let initial_value = 1.0 / width as f64;
+    Matrix::from_fn(topic_count, vocab_size, |col, row| {
+        if row < width {
+            // Horizontal topics
+            if col / width == row {
+                initial_value
+            } else {
+                0.0
+            }
+        } else {
+            //Vertical topics
+            if col % width == (row - width) {
+                initial_value
+            } else {
+                0.0
+            }
+        }
+    })
+}
+
+/// Samples `count` times from a dirichlet distribution with alpha as given and
+/// beta 1.0.
+fn get_dirichlet(count: usize, alpha: f64) -> Vector<f64> {
+    let mut rng = thread_rng();
+    let g_dist = gamma::Gamma::new(alpha, 1.0);
+    let result = Vector::from_fn(count, |_| {
+        g_dist.ind_sample(&mut rng)
+    });
+    let sum = result.sum();
+    result / sum
+}
+
+/// Generates a document based on a word distributiion as given.  The topics are randomly sampled
+/// from a dirichlet distribution and then the word sampled from the selected topic.
+fn generate_document(word_distribution: &Matrix<f64>, topic_count:usize, vocab_size: usize, document_length: usize, alpha: f64) -> Vec<usize> {
+    let mut document = vec![0; vocab_size];
+    let topic_distribution = get_dirichlet(topic_count, alpha);
+    for _ in 0..document_length {
+        let topic = choose_from(&topic_distribution);
+        let word = choose_from(&word_distribution.row(topic).into());
+        document[word] += 1;
+    }
+    document
+}
+
+/// Generate a collection of documents based on the word distribution
+fn generate_documents(word_distribution: &Matrix<f64>, topic_count: usize, vocab_size: usize, document_count: usize, document_length: usize, alpha: f64) -> Matrix<usize> {
+    let mut documents = Vec::with_capacity(vocab_size * document_count);
+    for _ in 0..document_count {
+        documents.append(&mut generate_document(word_distribution, topic_count, vocab_size, document_length, alpha));
+    }
+    Matrix::new(document_count, vocab_size, documents)
+}
+
+/// Displays the distrbution of words to a topic as a square graph
+fn display_topic(topic: &Vector<f64>, width: usize) {
+    let max = topic.iter().fold(0.0, |a, b|{
+         if a > *b {
+            a
+        } else {
+            *b
+        }
+    });
+    for (index, element) in topic.iter().enumerate() {
+        let col = index % width;
+        let out = element / max * 9.0;
+        print!("{}", out as usize);
+        if col == width - 1 {
+            print!("\n");
+        }
+    }
+}
+
+/// Chooses from a vector of probailities.
+fn choose_from(probability: &Vector<f64>) -> usize {
+    let mut rng = thread_rng();
+    let selection:f64 = rng.gen_range(0.0, 1.0);
+    let mut total:f64 = 0.0;
+    for (index, p) in probability.iter().enumerate() {
+        total += *p;
+        if total >= selection {
+            return index;
+        }
+    }
+    return probability.size() - 1;
+}
+
+pub fn main() {
+    let topic_count = 10;
+    let document_length = 100;
+    let document_count = 500;
+    let alpha = 0.1;
+    let width = topic_count / 2;
+    let vocab_count = width * width;
+    println!("Generating documents");
+    let word_distribution = generate_word_distribution(topic_count);
+    let input = generate_documents(&word_distribution, topic_count, vocab_count, document_count, document_length, alpha);
+    let lda = LDA::new(topic_count, alpha, 0.1);
+    println!("Predicting");
+    let result =  lda.predict(&(input, 30)).unwrap();
+    let dist = result.phi();
+    println!("Prediction completed");
+    for (topic, row) in dist.row_iter().enumerate() {
+        println!("\nTopic {}", topic);
+        display_topic(&row.into(), width);
+    }
+
+}
--- a/src/analysis/score.rs
+++ b/src/analysis/score.rs
@@ -44,9 +44,24 @@ pub fn accuracy<I>(outputs: I, targets: I) -> f64
     correct as f64 / len
 }
 
+
 /// Returns the fraction of outputs rows which match their target.
-pub fn row_accuracy(outputs: &Matrix<f64>, targets: &Matrix<f64>) -> f64 {
-    accuracy(outputs.iter_rows(), targets.iter_rows())
+pub fn row_accuracy<T: PartialEq>(outputs: &Matrix<T>, targets: &Matrix<T>) -> f64 {
+
+    assert!(outputs.rows() == targets.rows());
+    let len = outputs.rows() as f64;
+
+    let correct = outputs.row_iter()
+                         .zip(targets.row_iter())
+                         .filter(|&(ref x, ref y)| x.raw_slice()
+                                                    .iter()
+                                                    .zip(y.raw_slice())
+                                                    .all(|(v1, v2)| v1 == v2))
+                         .count();
+    correct as f64 / len
+
+    // Row doesn't impl PartialEq
+    // accuracy(outputs.row_iter(), targets.row_iter())
 }
 
 /// Returns the precision score for 2 class classification.
@@ -212,7 +227,8 @@ pub fn neg_mean_squared_error(outputs: &Matrix<f64>, targets: &Matrix<f64>) -> f
 #[cfg(test)]
 mod tests {
     use linalg::Matrix;
-    use super::{accuracy, precision, recall, f1, neg_mean_squared_error};
+    use super::{accuracy, precision, recall, f1, neg_mean_squared_error, row_accuracy};
+
 
     #[test]
     fn test_accuracy() {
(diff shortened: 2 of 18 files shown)


dyule:
Just realized I called it linear dirichlet allocation in the title.  It's Latent Dirichlet Allocation.  I constantly get that backwards, but it's right in the code.

AtheMathmo:
Thank you for this! I've been hoping to make LDA a part of rusty-machine for a while now :).

Sadly I'm no expert on LDA and I am particularly busy at the moment so I'll need a little time to digest this PR. Can you poke me if I still haven't reviewed this by the end of the week?

dyule:
If it helps, the only two files that are from LDA and not the rulinalg bump are `examples/lda_gen.rs` and `src/learning/lda.rs`.  I don't believe I changed anything else of importance.

dyule:
I've updated the code to be a little more efficient, and documented better.

AtheMathmo:
Thanks for your patience. I finally have some time to look through this and leave some comments. I'm still not too familiar with LDA but hopefully can provide some meaningful comments anyway.

examples/lda_gen.rs
```diff
@@ -0,0 +1,191 @@
+/// An example of how Latent Diriclhet Allocation (LDA) can be used.  This example begins by
+/// generating a distribution of words to categories.  This distribution is creatred so that
```
AtheMathmo: Typo here: `distribution is CREATED`.

examples/lda_gen.rs
```diff
@@ -0,0 +1,191 @@
...
+
+use rand::{thread_rng, Rng};
+use rand::distributions::{gamma, IndependentSample};
+
+use std::cmp::max;
+
+/// Given `topic_count` topics, this function will create a distrbution of words for each
+/// topic.  For simplicity, this function assumes that the total number of words in the corpus
+/// will be `(topic_count / 2)^2`.
+fn generate_word_distribution(topic_count: usize) -> Matrix<f64> {
```
AtheMathmo: Minor point: Potential issue here with assuming `topic_count` is divisible by 2. From what I can tell this shouldn't break anything but I wanted to draw you attention to it regardless.

src/analysis/score.rs
```diff
@@ -44,9 +44,24 @@ pub fn accuracy<I>(outputs: I, targets: I) -> f64
     correct as f64 / len
 }
 
+
 /// Returns the fraction of outputs rows which match their target.
-pub fn row_accuracy(outputs: &Matrix<f64>, targets: &Matrix<f64>) -> f64 {
-    accuracy(outputs.iter_rows(), targets.iter_rows())
+pub fn row_accuracy<T: PartialEq>(outputs: &Matrix<T>, targets: &Matrix<T>) -> f64 {
+
+    assert!(outputs.rows() == targets.rows());
```
AtheMathmo: I'm a little confused by this change. I'm guessing it comes from the `rulinalg-0.4` PR?

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
+//! Latent Diriclhet Allocation Module
```
AtheMathmo: Minor: Type here, `DIRICHLET`.

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
...
+//!
+//! Contains an implementation of Latent Diriclhet Allocation (LDA) using
```
AtheMathmo: And again here

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
...
+//! [Gibbs Sampling](http://psiexp.ss.uci.edu/research/papers/sciencetopics.pdf)
+//!
+//! LDA is typically used for textual analysis.  It assumes that a topic can be modeled
+//! as a distribution of words, and that a document can be modeled as a distribution of
+//! topics.  Thus, the output is these distributions on the input documents.
+//!
+//! Gibbs sampling is a Morkov Chain Monto Carlo algorithm that iteratively approximates
+//! the above distributions.
+//!
+//! This module doesn't use any training.  It uses unsupervised learning to estimate
```
AtheMathmo: I find this statement a little confusing. I would argue that unsupervised learning algorithms still have a _training_ phase. However, during this training stage they are not given any target data.

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
...
+    topic_word_count: Matrix<f64>,
+    // The two vectors are simply used to reduce the amount of calculation that must be
+    // performed during the conditional probability step
+    topic_total_by_document: Vector<f64>,
+    word_total_by_topic: Vector<f64>,
+    alpha: f64,
+    beta: f64,
+}
+
+impl Default for LDA {
```
AtheMathmo: I'm unfamiliar with the algorithm and related literature, so please let me know if I'm being silly.

This seems a little arbitrary to me (the topic count in particular). We should at least document these values in the implementation, e.g.

```rust
/// Creates a new LDA with topic count = 10, ... etc.
impl Default for LDA { }
```

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
...
+            }
+            topics.push(document_topics);
+        }
+        (result, topics)
+    }
+
+    /// Find the distribution of words over topics.  This gives a matrix where the rows are
+    /// topics and the columns are words.  Each entry (topic, word) gives the probability of
+    /// word given topic.
+    pub fn phi(&self) -> Matrix<f64> {
```
AtheMathmo: This name seems somewhat dependent on related literature. I worry that unfamiliar users (like myself) would not know where to look for this function. Maybe it should be renamed `distr_of_words`?

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
...
+
+        // Normalize it so that it's a probability
+        probability /= probability.sum();
+        return probability;
+    }
+}
+
+impl UnSupModel<(Matrix<usize>, usize), LDAResult> for LDA {
+    /// Predict categories from the input matrix.
+        fn predict(&self, inputs: &(Matrix<usize>, usize)) -> LearningResult<LDAResult> {
```
AtheMathmo: I think that we need to be clearer about what these inputs are. I am assuming that the first tuple argument is the word counts per document, and the second is the number of iterations?

It would follow the rest of the library more closely if the iterations were specified at the model level, and this function simply took the `&Matrix<usize>`.

And one other small note, in order for docs to show up with rustdoc they must be placed on the outside of the impl block. e.g.

```rust
/// Doc here will be rendered
impl UnSupModel<..> ... {
    /// Docs here will not
    fn predict(...) -> ... { }
}
```

src/learning/lda.rs
```diff
@@ -0,0 +1,250 @@
...
+    }
+    return probability.size() - 1;
+}
+
+#[cfg(test)]
+mod test {
+    use super::{LDAResult, LDA};
+    use linalg::{Matrix, Vector};
+    #[test]
+    fn test_conditional_distribution() {
```
AtheMathmo: This is probably a result of me not knowing the algorithm very well, but this test does not look entirely convincing. How do you know that the conditional distribution returned here is indeed correct?

I don't have any better suggestions right now for improving the tests but I would feel more comfortable with stronger coverage for this algorithm.

src/analysis/score.rs
dyule: Indeed, I am pretty certain I took this one directly from the linalg bump PR

AtheMathmo: I've now merged a new linalg bump PR into master. I guess you'll want to rebase from that (or just C+P the relevant modules onto a new branch :) )

src/learning/lda.rs
dyule: LDA (in its original form) is in a sort of strange place with respect to machine learning, in that it's designed to classify only the input data, and nothing more.  In this sense, the training and prediction stage are actually identical.  One could certainly use the distribution that was discovered to classify additional documents, but for whatever reason, that's not how it was used.  I'll have more to say about this in a later comment.

src/learning/lda.rs
dyule: As far as I can tell, there aren't any sensible constant defaults for LDA.  I only added this default implementation because the other models seemed to have it.  I could drop it if you like (or document it better, of course).

src/learning/lda.rs
dyule: I agree that this is a confusing name.  My reason for keeping away from a simpler name was a) its use in the literature and b) I was a bit concerned it wouldn't be clear it was a conditional distribution (or the direction of the condition).  But I think with proper documentation, this is fine.

src/learning/lda.rs
dyule: I've been very unhappy with this function signature since I started.  It will certainly be more ergonomic to have them in the model, but my reasoning for keeping them out is that I don't think number of iterations is a property of the model.  However, I believe that my conception of the way your library was organized was a bit off, since in my brain, LDA was a sort of factory object that could generate multiple classifications for different input.  It seems pretty obvious now that this wasn't the case.

src/learning/lda.rs
dyule: I can look into testing a bit more, and I agree about the coverage.  This test is based on values I calculated basically by hand, and verified by a python implementation.  The present implementation of the algorithm really only has three functions of note: `predict` and `conditional_distribution` in `LDA` and `new` in `LDAResult`.  `predict` is basically the entire algorithm, and so is tested basically by the example.  It can't really by tested automatically, because it relies on random choice, and unless we want to make it possible to pass a source of randomness into a model, that can't really be controlled.  `new` has the same problem as `predict`, and `conditional_distribution` is already tested via these magic numbers.

I can certainly add in some tests of the same kind as some of the other learning algorithms, where I provide some basic input, and ensure that the algorithm completes without panicking.  I will include this in my updates.

dyule:
Thanks for taking the time to review this.  I very much appreciate all your comments.  I've responded to each of them with my thought process, but I will incorporate them into my changes.

For the example, I suppose I missed the listing of examples in the README, I will make sure to include it.  I've documented the file pretty extensively, since as you say, it's fairly involved.

The algorithm you've shown does seem to be an improvement, and one that I'd like to incorporate down the road.  My reasoning for using Gibbs sampling was a) it was the original approach and b) I actually implemented it for my own research, where I'm obligated to use that particular approach.  As I say, I'm happy to include it, but doing so will take some time that I possibly do not have at the moment.

As I mentioned in my inline comments, I misunderstood a few things about the organization of the crate.  I plan to move the main code from `predict` to `train` and make predict be a no-op.  However, looking at the paper you linked, and the scikit implementation, it seems like `predict` is very similar to scikit's `transform`.  That is, passing input data into `transform` results in a distribution of categories over documents.  I believe I could do this with the current algorithm, but it will be better once we're using online VB.

So, I will take some time to update the changes suggested, but leaving the central algorithm intact, then let you know.  When I have time, I'll look into changing to Online VB, for a faster and more flexible approach.

AtheMathmo:
I haven't read all your comments but from your summary I think your approach sounds great. Keep the core algorithm as it is and we can write up an issue for future work with VB once this has been merged.

>  it seems like `predict` is very similar to scikit's `transform`

Scikit uses both a `predict` and a `transform` function depending on the model. See their [GMM docs](http://scikit-learn.org/stable/modules/generated/sklearn.mixture.GaussianMixture.html). We also have both in [rusty-machine](https://athemathmo.github.io/rusty-machine/doc/rusty_machine/data/transforms/trait.Transformer.html) (though note that there have been some changes to this trait on the current master branch). If you think this will provide a better fit you could switch to this trait instead. Right now these traits are a little disorganized but I'm hoping to address that in a future release.

src/learning/lda.rs
AtheMathmo: I think it is good to have a default implementation so people can easily get off the ground. But yes it would be good to add documentation for this and explicitly state the parameter choices made.

src/learning/lda.rs
AtheMathmo: > This test is based on values I calculated basically by hand, and verified by a python implementation.

I think this is fine but you should note this somewhere near the test with a reference to any implementation you used if possible. It makes them a little more convincing to people reading over the code.

>  unless we want to make it possible to pass a source of randomness into a model

This is actually a good idea. See #138 . The support isn't there in the library yet but if you can get something working in the mean time it would be helpful for later. If we want to make changes later we'd like to be able to verify that the algorithm is still doing (close-to) the same thing.

src/learning/lda.rs
AtheMathmo: This isn't your fault, I'm quite unhappy with the structure of the current traits. They worked well for the first few algorithms I implemented but are a bit of a stretch now for unsupervised models.

Here you have sort of emulated what I want the traits to look like. One trait should be used to train a model using input data and this should output a model that can be used predictively. Like how your `LDA` model produces an `LDAResult`. The new `Transformer` trait works in this way.

dyule:
The `Transform` trait does in fact seem like an excellent fit.  I have only one thought.  Currently, the parameterization of `TransformFitter` uses `T` to indicate a `Transformer` whereas the parameterization of `Transformer` uses `T` to indicate some kind of input type.  It's a small thing, but it did cause me some confusion when first reading the code.

AtheMathmo:
> Also, transformer current requires the input and output of the transform to be of the same type

Ah yes, that is true. I'm not sure whether we want to lift this restriction as future changes to the `UnSupModel` should give the behaviour you want. I'll have to think about things a little more...

dyule:
So, I've made the changes promised.  I've implemented LDA as a Transformer, which involved allowing them to have different input and output types.  I'm happy to switch it back to being a learning model if that's where you think it fits best.

What it comes down to is the semantics of a `Transformer`, and where LDA fits into that idea.  Currently, LDA has much more in common with the other learning models than with the other transformers, which all have to do with pre-processing the input data, but I'm not sure what your long term intention is there.

patrickmesana:
Don't you mean **Latent** Dirichlet Allocation?

rohitjoshi:
👍 waiting for merge

zackmdavis:
>  waiting for merge

(Deputy acting co-maintainer here, but time-crunched at the moment; have made a note to take a look early next week)

zackmdavis:
(or maybe this week)

PR #173, opened 2017-01-24
Implemented the poisson, binomial and multinomial distributions

dyule:
Basically, as it says.  It also implements the gamma function for Rust, since I could not find any other implementations.  After some refining here, I may try to submit that part to the main Rust library, since they link it from C when available (i.e. not on Windows).

A great deal of the sampling code is pretty dense, and algorithmic.  This is because it is based on previous implementations, which are themselves based directly on some academic papers with obtuse notation (references and such are noted in the comments)

Let me know what you think, and I can continue to iterate.

### DIFF
--- a/src/stats/dist/binomial.rs
+++ b/src/stats/dist/binomial.rs
@@ -1,0 +1,330 @@
+//! Binomial distribution module.
+//!
+//! Contains extension methods for the Binomial struct
+//! found in the rand crate. This is provided through
+//! traits added within the containing stats module.
+
+use stats::dist::Distribution;
+use rand::Rng;
+use rand::distributions::{Sample, IndependentSample};
+
+use std::cmp::min;
+
+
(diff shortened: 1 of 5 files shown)


AtheMathmo:
I am happy to review and merge this, I can see you have put a lot of effort in and it looks great!

But I had planned on removing this section of rusty-machine at some point. I discovered [statrs](https://github.com/boxtown/statrs) which offers much of this functionality in a dedicated fashion. I plan to incorporate this into rusty-machine rather than reimplementing everything.

With that said I haven't spent too much time thinking about this, what is your opinion?

dyule:
I hadn't seen statrs when I started on this, or I wouldn't have.  I looked for something similar, but I guess my search skills were weak.  Anyways, I can't see much value in including this here when it's already implemented much cleaner elsewhere.

PR #178, opened 2017-04-25
Adding text vectorization

z1mvader:
This is oriented to https://github.com/AtheMathmo/rusty-machine/issues/177 . It adds the frequency vectorizer and a simple tokenizer

### DIFF
--- a/src/data/tokenizers/mod.rs
+++ b/src/data/tokenizers/mod.rs
@@ -1,0 +1,19 @@
+//! The Tokenizer module
+//!
+//! This module contains traits used to tokenize text.
+//! It also reexports these `Tokenizer`s from child modules.
+//!
+//! The `Tokenizer`s are intended to be used with the text
+//! `Vectorizer`
+
+mod naive;
+
+pub use self::naive::NaiveTokenizer;
+
+use learning::LearningResult;
+
+/// A trait used to construct Tokenizers
+pub trait Tokenizer {
+    /// Tokenize the inputs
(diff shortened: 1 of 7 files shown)


z1mvader:
@AtheMathmo any concerns?

AtheMathmo:
Sorry for the delay in reviewing this, it's a particularly busy time for me right now...

At a cursory glance most the details here look good. I'm a little concerned about the addition of the `Vectorizer` trait. I feel that a modification to the `Transformer` trait might be more appropriate. But honestly I'd need to review the code properly to be sure.

I'll try to get around to this by the end of the week. Please do poke me again if I haven't given any feedback by then.

z1mvader:
AtheMathmo thanks for your answer given that you are busy right now!
If you want I can modify the `Transformer` trait and replacing all the implementations of it with the correct generics. I have a few concerns about this, the most important is about the fact that loading strings with the text as owned stuff. It would be better to load it from slices to save unnecessary allocations.

k0pernicus:
It seems that @ghost has disappeared...
Is a new text vectorization feature has been accepted since this PR @AtheMathmo ?

zackmdavis:
@k0pernicus The deactivation of the author's GitHub account is probably unrelated to this project and therefore doesn't necessarily stop us from using his patches ...

(Rusty-machine has been very quiet lately—albeit crucially, not entirely dead!—while Athe has been busy with grad school and I've only managed to pitch in some review effort with spare cycles here and there; so text vectorization is definitely on the table, just without any timeline promises.)

PR #179, opened 2017-07-17
Linear Regressor now returns error instead of panicking

mark-blum:
Before Linear Regressor would panic on ill-conditioned data, now it actually returns it as a result.

### DIFF
--- a/src/learning/lin_reg.rs
+++ b/src/learning/lin_reg.rs
@@ -87,8 +87,7 @@ impl SupModel<Matrix<f64>, Vector<f64>> for LinRegressor {
         let full_inputs = ones.hcat(inputs);
 
         let xt = full_inputs.transpose();
-        self.parameters = Some((&xt * full_inputs).solve(&xt * targets)
-                                                  .expect("Unable to solve linear equation."));
+        self.parameters = Some((&xt * full_inputs).solve(&xt * targets)?);
         Ok(())
     }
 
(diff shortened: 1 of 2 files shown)


mark-blum:
So in this commit I added a test so that the Linear Regressor actually returns as error when given e.g. no data, but I've noticed that it isn't consistent. Somehow, it randomly decides whether it will pass or fail

AtheMathmo:
The PR looks good to me but I'm a little confused by your comment. I believe that the solver should be deterministic (up to hardware noise I guess?) But you find that sometimes the solver will claim to be able to deal with the empty matrices?

It may be a bug with the solver not handling the empty matrices correctly (though that doesn't explain the funky behaviour you describe). If so then maybe to get this merged it would make more sense to write a test with ill-conditioned data to ensure the solver fails.

mark-blum:
What I mean by the comment is that in running `cargo test` multiple times, the test that I wrote sometimes passes (as in training a linear regressor with an empty matrix and empty vector returns an `Err(...)`) but sometimes it returns an `Ok(...)`. I haven't looked into the underlying code to see what could be causing this but it's certainly strange that it isn't deterministic

AtheMathmo:
I agree - in this case especially. I think that empty matrices should be handled explicitly in rulinalg. I've added the bug tag though I need to run this myself to confirm.

mark-blum:
Off the top of my head I would guess either some race condition with threads, though I don't think there are threads here, or something weird is happening with a static object.

zackmdavis:
Nondeterminism confirmed (and filed as #183), but there's no reason not to merge this pull request: the bug is already lurking somewhere on master; the new test merely exposes it (and returning a result instead of panicking is better in any case).

zackmdavis:
@mark-blum thanks! :sparkling_heart:

PR #190, opened 2018-01-05
Nesterov Momentum

cfeitong:
add Nesterov momentum for SGD


zackmdavis:
thanks! :sparkling_heart:

AtheMathmo:
Just want to add a quick note - I'm not sure we would want Nesterov momentum to be default. It is great to have it as an option but I expect many users will be surprised if this is the default behaviour. This is especially important as from what I can tell they can no longer use SGD with Nesterov momentum?

Perhaps add an `Option` to the constructor?

PR #194, opened 2018-03-31
Fix link to API documentation in contributing.md

moberst:
The link in the contributing file points to a dead link, which is inconsistent with the link in the main README.md file, which seems to work.  So just making a simple tweak to the CONTRIBUTING.md file =)


zackmdavis:
@moberst thanks!

PR #198, opened 2018-11-30
Autoformat crate

tiziano88:
This was done by running the `cargo fmt` command on the crate.


tiziano88:
Thanks for the reply! The format PR was in fact a prerequisite for more refactorings that I would like to send to you soon, if there is interest. I think you should check the code, or run the same command yourself and check there are no extraneous diffs, it should be a quick one. Anyway let me know what you prefer, I would like it for rusty-machine to become the de facto ML crate in Rust!

PR #205, opened 2019-05-06
Update rand to 0.6 and add wasm-bindgen feature enabling corresponding rand feature

DenialAdams:
This enables to me to use rusty-machine inside of WASM.

Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,5 +17,5 @@ datasets = []
 
 [dependencies]
 num = { version = "0.1.41", default-features = false }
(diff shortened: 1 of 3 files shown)


DenialAdams:
Ah, I didn't try compiling with features, so some more updates have to be made. I'll try to tackle that in a bit.

DenialAdams:
@AtheMathmo took me a long time, but this should work for all features now. This updates rand to 0.7, which should be a nice boon for users, and allows rusty-machine to be used in WASM

PR #206, opened 2019-05-26
Replace recursive dbscan with iterative

sjeohp:
Prevent recursion stack overflow and make fewer allocations on large
dbscan clusters.


AtheMathmo:
I'd like to merge this after the conflict I've introduced is resolved.

Issue #200, opened 2018-12-06
Linear regression model panics if unable to solve linear equation.


MichaelMauderer:
The [`LinRegressor.train`](https://github.com/AtheMathmo/rusty-machine/blob/e23ae6deff6f8ec3ec0e761df4a45d32535c039a/src/learning/lin_reg.rs#L91) method panics if it is unable to solve the given linear equation. It would be better if it returned a `LearningResult` indicating the failure.

MichaelMauderer:
Actually, it seems this is fixed in [the master branch ](https://github.com/AtheMathmo/rusty-machine/blob/f43da8d7696ff2c8f9bb7d2d049ed06c4bfaaacf/src/learning/lin_reg.rs#L90)but hasn't been released yet.

PR #208, opened 2019-10-30
Use `slice::iter` instead of `into_iter` to avoid future breakage

martin-fink:
Use `slice::iter` instead of `into_iter` to avoid future breakage

`an_array.into_iter()` currently just works because of the autoref
feature, which then calls `<[T] as IntoIterator>::into_iter`. But
in the future, arrays will implement `IntoIterator`, too. In order
to avoid problems in the future, the call is replaced by `iter()`
which is shorter and more explicit.

A crater run showed that your crate is affected by a potential future 
change. See https://github.com/rust-lang/rust/pull/65819 for more information.

### DIFF
--- a/src/learning/naive_bayes.rs
+++ b/src/learning/naive_bayes.rs
@@ -186,7 +186,7 @@ impl<T: Distribution> NaiveBayes<T> {
 
     fn find_class(row: &[f64]) -> LearningResult<usize> {
         // Find the `1` entry in the row
-        for (idx, r) in row.into_iter().enumerate() {
+        for (idx, r) in row.iter().enumerate() {
             if *r == 1f64 {
                 return Ok(idx);
             }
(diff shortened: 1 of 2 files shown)


tafia:
i think the `into_iter` are legit on the tests (vecs must be consumed as they are not used later).

martin-fink:
You are right, thanks!

AtheMathmo:
I'm afraid I've been away from the rust ecosystem to understand this issue without digging deeper. Should this PR be closed following @tafia 's comment?

LukasKalbertodt:
Hi there! This PR is indeed not necessary. The code changed in this PR does not lead to breakage as `into_iter` is called on slices, not arrays. While the change is not bad (`iter()` is shorter and does the same), it's not necessary.

`rusty-machine` regressed in our tests because of the use of `matrix!` from `rulinalg` ([build log](https://crater-reports.s3.amazonaws.com/pr-65819/try%23b34fa46cc6fa5d46c6adcb041ebd28c7f7a374f3/reg/rusty-machine-0.5.4/log.txt)). There is already an open PR fixing the incorrect code in `rulinalg`: https://github.com/AtheMathmo/rulinalg/pull/200
As this seems to be also your repository, it should be fine with getting it merged?

@AtheMathmo In short: close this PR, but merge [this one](https://github.com/AtheMathmo/rulinalg/pull/200).

Also: please consider [Rust Bus](https://users.rust-lang.org/t/bus-factor-1-for-crates/17046) to aid with maintaining your two crates. They do maintaining for you and will find a new maintainer.

